PR #47, opened 2015-09-23
Fix bug and add duplicated phone number check.

wjcheers:
1. fix bug:
candidate work phone filter is not working
2. add function:
check phone number is duplicated when add a candidate.
3. fix correct index name:
in lib/CATSUtility.php


RussH:
Looks good - will merge then test on my test laptop.. Thank you!!

PR #61, opened 2016-03-09
Update suggest.js

theSlyest:
Line 216. Since it comes from the database, nameNodeValue may contain single quotes. And if it does, it obviously causes a bug if the quote is not escaped, so I just escaped it.


libregeek:
I think we could merge this. But in the long-term I would suggest integrating something like jQuery so that we can concentrate on ATS features.

RussH:
okay, will merge!




----
Russ

PR #66, opened 2016-05-19
Added grave accents, backticks, to database schema name

aaam:
When an existing database schema uses a reserved name, like numbers only, backticks prevent errors.


RussH:
jsut for completeness - analysed on Codacy. all looks good.

PR #73, opened 2016-06-23
Ldap module improvements

skrchnavy:
Included support for combined authentication against LDAP and SQL.
Configuration tested with active directory.
More details are described in INSTALL.md


RussH:
@libregeek can you review and comment / accept as you're the owner of all things LDAP!

Issue #58, opened 2016-02-26
Use fluid width pages


libregeek:
OpenCATS uses fixed width pages (1024x768 resolution) which doesn't looks good in wide screens. The fixed width should be changed to fluid width so that the application utilized available screen space and reduces vertical scrolling.
![selection_410](https://cloud.githubusercontent.com/assets/145110/13346128/24536430-dc8a-11e5-8e42-19290f52c230.png)

mlespiau:
Hi @libregeek, would you consider this issue fixed since my last PR?

amaisonneuve:
I took a look and the changes are clean I have yet to merge into my repo, but I think it should be safe.

Cheers,
Alain

PR #74, opened 2016-06-28
Flexible layout

mlespiau:
Remove fixed width layout


RussH:
@libregeek can you review please?

mlespiau:
@libregeek, @RussH  would appreciate the review, I'm getting confused with all the open branches as I would like to use them all :)

skrchnavy:
expects it closes #58

Issue #42, opened 2015-08-25
Decoding of UTF-8 characters mangles Calendar entries


pes-soft:
When there are special characters ( non-latin ) used in Calendar entries ( as Calendar entry title or name of the person involved ), these entries are not decoded from event strings correctly by JavaScript and are displayed mangles ( as UTF-8 string in ASCII ).

To fix this behaviour, it is required to replace `unescape` with `decodeURIComponent` within *modules/calendar/Calendar.js* in function *calendarDataPopulateString*.

skrchnavy:
PR #84 shall fix this issue

PR #84, opened 2016-07-08
Fixed issues with UTF8

skrchnavy:
This is fix of OpenCATS/issues#42 and also few other display issue related to change DB to UTF8.

Tested with db in `utf8_general_ci` and `utf8_unicode_ci`


RussH:
closes #42

PR #89, opened 2016-07-17
Wait for mysql and apache before running tests in travis

mlespiau:
This is an important change in terms of continuous integration. 
As for now, I'd been using a simple sleep to wait for the testing environment to start, but even though I set a large time such as 12 seconds, on travis sometimes the servers takes longer than that and you get a failure on your tests and need to read the logs before being able to understand that it's due to the environment not being fully started. 

In order for this not to be confusing to other people willing to commit to the repository, I've added a script that waits for the web server, database and dump to be ready before running the tests with a timeout of 30 seconds. 

So, now travis will run tests as soon as the environment is ready or fail after waiting for 30 seconds for the environment to start.


RussH:
Definitely approved!!

RussH:
Definitely approved!!

mlespiau:
:)

PR #90, opened 2016-07-23
docker - added execute permission for installing composr

skrchnavy:
issue was observed when running docker on windows, permission denied were reported during installation.

### DIFF
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -22,7 +22,7 @@ RUN cd ~ && \
   make install
 RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ && docker-php-ext-install -j5 mysql gd ldap soap zip
 ADD scripts/install-composer.sh /opt/install-composer.sh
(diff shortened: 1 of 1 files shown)


mlespiau:
Hi @skrchnavy, can you provide more details on your windows setup? The file "install-composer.sh" has executable permissions and git configuration has core.filemode = true, which should make those permissions to be retained on windows systems. 

From [git documentation](https://git-scm.com/docs/git-config):
> A repository, however, may be on a filesystem that handles the filemode correctly, and this variable is set to true when created, but later may be made accessible from another environment that loses the filemode (e.g. exporting ext4 via CIFS mount, visiting a Cygwin created repository with Git for Windows or Eclipse). In such a case it may be necessary to set this variable to false. See git-update-index[1].

skrchnavy:
It is windows 7 system, sourcetreeapp with git version 1.8.3.msysgit.0
I checked git attributes of the file, and it has execute permission set and auto_crlf = true.
I think this is docker issue when it ads file using `ADD scripts/install-composer.sh /opt/install-composer.sh`, it probably doesn't transfer permissions and also owner is not clear.
I checked some DockerFiles where are distributed files and they are using chmod/chown commands to properly setup the container.

PR #91, opened 2016-07-26
Security improvements using secured objects

skrchnavy:
As proposed in this topic in forum: https://groups.google.com/forum/#!topic/opencats-dev/MgZPtihEJ_M this is implementation of improved access level check based on secured objects.

It is backward compatible, so when no permissions defined per object, access level of defined user is used.
Changes made:
- changed implementation of Session::getAccessLevel
  - _Open question: shall we treat default access level for user to be maximum access level?_
- checks of access level in modules moved to handle request
- some public modules functions made private (not used out of class)
- menu (tabs/subtabs) access level checks also for secured object name (documented in function).

What is not finished / needs further development
- access level check not added into all places (only where it was used)
- secured object names corrections, simplification
- reading permissions / access levels (ACL) from a configuration file
- tabs/subtabs to be displayed based on permissions also for read only access
- automatic tests for different configuration of permissions


RussH:
Hi Sveto, could you add in some unit tests to travis to check that this functionality works?

skrchnavy:
We will try to add some tests, I understand it is quite big change, so first goal is to prepare environment for the functionality and keep the system working as before.
The most risky is module settings (administration) where were used realAccessLevel instead of accessLevel. Other modules are currently 1:1 mapping.

ajax/getPipelineJobOrder.php
```diff
@@ -276,7 +276,7 @@ function printSortLink($field, $delimiter = "'", $changeDirection = true)
             <td valign="top"><?php echo($pipelinesData['lastActivity']); ?></td>
 <?php if (!$isPopup): ?>
             <td align="center" nowrap="nowrap">
-                <?php if ($_SESSION['CATS']->getAccessLevel() >= ACCESS_LEVEL_EDIT && !$_SESSION['CATS']->hasUserCategory('sourcer')): ?>
+                <?php if ($_SESSION['CATS']->getAccessLevel('pipelines.screening') >= ACCESS_LEVEL_EDIT && !$_SESSION['CATS']->hasUserCategory('sourcer')): ?>
```
mlespiau: `'pipelines.screening'` is a magic string here. This is no good for maintenance, all `$securedObjectName` should be tracked as constants, something like: 

`
class SecuredObjects {
    const PIPELINES_SCREENING = 'pipelines.screening';
}
`
This comment applies to all magic strings being passed to `getAccessLevel` method.

lib/ACL.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+/*
+ * OpenCATS
+ * ACL Library
+ * author:
```
mlespiau: This comments seems rather incomplete. I guess we should standardise on what do we want to place at the top of new PHP class files. @RussH?

lib/JobOrders.php
```diff
@@ -1257,7 +1257,7 @@ public function __construct($instanceName, $siteID, $parameters, $misc)
     public function getSQL($selectSQL, $joinSQL, $whereSQL, $havingSQL, $orderSQL, $limitSQL, $distinct = '')
     {
         // FIXME: Factor out Session dependency.
-        if ($_SESSION['CATS']->isLoggedIn() && $_SESSION['CATS']->getAccessLevel() < ACCESS_LEVEL_MULTI_SA)
+        if ($_SESSION['CATS']->isLoggedIn() && $_SESSION['CATS']->getAccessLevel('') < ACCESS_LEVEL_MULTI_SA)
```
mlespiau: `getAccessLevel('')` it's not possible to understand what's the effect of calling getAccessLevel with an empty string. The only way I could figure out is by reading the implementation. 

I think this API can be improved. What's the intention here?

lib/Session.php
```diff
@@ -396,8 +398,37 @@ public function isDateDMY()
     }
 
     // FIXME: Document me!
-    public function getAccessLevel()
+    public function getAccessLevel($securedObjectName)
     {
+        $ACL = ACL::getInstance();
+        $permissions = $ACL->permissions;
```
mlespiau: Encapsulation for ACL instance is broken.
$permissions array should not be publicly exposed. See below a few suggestions to improve it.

lib/Session.php
```diff
@@ -396,8 +398,37 @@ public function isDateDMY()
...
+
+        if(empty($permissions) || isset($permissions) == false || isset($securedObjectName) == false)
```
mlespiau: The use of `isset($permissions) == false` and `isset($securedObjectName) == false` in this code are conceptually wrong:

- the `$permissions` variable is defined a few lines above so this is always going to be "set"
- the `$securedObjectName` variable is a parameter to this method so this is always going to be "set".

The `empty($permissions)` should be encapsulated inside a new method in the ACL class. 
There is a missing whitespace between the if and parethensis.

The resulting if could be:
`if ($ACL->isEmpty())`

And in the $ACL class could have the following method:
`
public function isEmpty()
{
    return empty($this->permissions);
}
`

lib/ACL.php
RussH: the intention was always that any new files (or significantly altered
original files) would be completely open-source. The existing files are CPL
- a modified version of the Mozilla Public License. Therefore it would be
simplest to keep new files as MPL;
https://www.mozilla.org/en-US/MPL/headers/

/* This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */

OR

# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.

OR




With 'original' or 'significantly original'  files using;

* The Original Code is "CATS Standard Edition".
*
* The Initial Developer of the Original Code is Cognizo Technologies, Inc.
* Portions created by the Initial Developer are Copyright (C) 2005 - 2007
* (or from the year in which this file was created to the year 2007) by
* Cognizo Technologies, Inc. All Rights Reserved.

- I'll update the License statement in the root with some words to the
effect that


This application is covered by two licenses.
1) Mozilla Public License 2.0 for all new and significantly altered code.
2) The Original Code is "CATS Public License Version 1.1a"  (a modified
Mozilla Public License)



----
Russ

lib/Session.php
```diff
@@ -396,8 +398,37 @@ public function isDateDMY()
...
+        {
+            return $this->_accessLevel;
+        }
+
+        $userCategory = $this->getUserCategories()[0];
+        if( isset($userCategory) == false)
+        {
```
mlespiau: Encapsulation is broken as `getAccessLevel` needs to know the internals about `getUserCategories`.  
Also makes the code less readable. 

What's the intention here? 

Looks like this code is always using the first category no matter how many are there. 

instead of `
$userCategory = $this->getUserCategories()[0];
if( isset($userCategory) == false)`

The could could be `
$userCategory = $this->getFirstUserCategory();
if(!$userCategory)
`

lib/Session.php
```diff
@@ -396,8 +398,37 @@ public function isDateDMY()
...
+            return $this->_accessLevel;
+        }
+
+        if(isset($permissions[$userCategory][$securedObjectName]))
+        {
+            return $permissions[$userCategory][$securedObjectName];
+        }
```
mlespiau: To me, these checks should be responsibility of the ACL class. It should have a method like: getAccessFor($userCategory, $securedObjectName)`

lib/Session.php
```diff
@@ -396,8 +398,37 @@ public function isDateDMY()
...
+        else
+        {
+            while(($pos = strrpos($securedObjectName, ".")) !== false)
+            {
+                $securedObjectName = substr($securedObjectName, 0, $pos);
+                if(isset($permissions[$userCategory][$securedObjectName]))
+                {
+                    return $permissions[$userCategory][$securedObjectName];
+                }
+            }
```
mlespiau: Same here.

lib/Template.php
```diff
@@ -127,6 +127,14 @@ public function display($template)
 
         echo($html);
     }
+
+    /**
+     *  TODO: Document me.
+     */
```
mlespiau: Pending TODO

lib/TemplateUtility.php
```diff
@@ -705,7 +716,14 @@ public static function printTabs($active, $subActive = '', $forceHighlight = '')
...
                         $al = substr($link, $alPosition + 4);
-                        if ($_SESSION['CATS']->getAccessLevel() >= $al ||
+                        $soPosition = strpos($al, "@");
+                        $soName = '';
+                        if( $soPosition !== false )
+                        {
+                            $soName = substr($al, $soPosition + 1);
+                            $al = substr($al, 0, $soPosition);
+                        }
+                        if ($_SESSION['CATS']->getAccessLevel($soName) >= $al ||
```
mlespiau: Duplicated code.

mlespiau:
Hi @skrchnavy, it's a great change the one that you're doing here. I've started the review but could not complete it yet. 

Nevertheless, I have a question, it seems that you are somehow allowing previous and new implementation to co-exist. I think the code to handle it that way is far more complex than just moving to the new implementation and setting up the permissions in the new ACL way. I would prefer that approach, what are your thoughts?

lib/JobOrders.php
skrchnavy: empty secured object means root object. It is intended also to comment regarding magic constants, this shall be replaced properly.

lib/Session.php
skrchnavy: Then it will be needed also to send access level of user to ACL or to make some checks if ACL is not configured to return access level of user. This is necessary for backward compatibility.

ajax/getPipelineJobOrder.php
skrchnavy: The goal is to define secured objects names using convention <module>.<action>. Each module shall define own set of securedObjectNames. Inside module it is quite clear and easy to maintain. Unfortunately these ajax sources are implemented out of modules.

lib/ACL.php
skrchnavy: I agree to have common header for new code. Not all sources have the header.

skrchnavy:
The main goal of this pull request is to prepare environment for further improvements of security. For now is to add names of secured objects without change of current functionality and without schema change.

In next, we want to maintain better security for different roles in recruitment process
* recruiter is responsible for candidates
* hiring manager(s) are responsible for job orders and selection of proper candidates (in pipeline for job order)
* HR department to see candidates and job orders and to be able to manage candidates when recruiter not present.

Also there shall be added possibility to disable read only access to some tabs/subtabs.
And also some roles can have access only to objects they owns (this is in analysis phase how to achieve this).

lib/TemplateUtility.php
skrchnavy: not clear why do you think it is duplicate

mlespiau:
The feature you're building seems great but I would like to make sure we can keep improving the code base maintainability. Would you like to do a short call to discuss your changes? 
Also it would be great to have some documentation about this planned feature. 

I'm happy to support and provide in-depth reviews for this code.

skrchnavy:
We will make updates in code to address some of your comments to reduce the list of questions and then we can make a call. 
I am not php developer so if you have ideas of improvements, please make changes directly in the code or in new branch from this branch.
I could extract some problematic code from this branch to new branch and to add there some further changes we plan.
I would like to finalise this branch to be able to use the approach in other features we are working on to avoid big fork from main code.

RussH:
@mlespiau I'll rely on you to review and merge this pull when you're happy!

skrchnavy:
regarding brief doc, I also started few days ago topic at dev forum: 
https://groups.google.com/forum/#!topic/opencats-dev/MgZPtihEJ_M

lib/JobOrders.php
skrchnavy: comment addressed in commit 4c48ff3c0068952eda5dd26beae645be611da1cb

lib/Session.php
skrchnavy: this comment addressed in commit 20e0d83688289e791fa55bdc1edb9e9f2e23876b

lib/Session.php
skrchnavy: this comment addressed in commit 1f441e88f1e4b36ec8f676bb39e8ff6351d33c13

lib/Session.php
skrchnavy: this comment addressed in comment 20e0d83688289e791fa55bdc1edb9e9f2e23876b

lib/Session.php
skrchnavy: this comment addressed in commit 20e0d83688289e791fa55bdc1edb9e9f2e23876b

lib/Template.php
skrchnavy: comment addressed in commit 9d6845a06d5c75a7d5dd0ee66fe052fc56994e62

RussH:
Hi both, did you agree that this change is good, or still subject to review?

skrchnavy:
Hi @RussH , for now it is still under development, we are working on tests in master branch and when tests there and in this branch pass we can merge it.

mlespiau:
Hi @skrchnavy, @amaisonneuve and @RussH. 

I would like to setup a meeting/video conference to discuss this code and it's implementation. I love the fact that @skrchnavy is working hard on this functionality but I'm concerned in proceeding with merging this code to the trunk. I guess in a meeting we could reach out to a good acceptance criteria that might yield better results:
* lack of previous tests before all the changes (super hard to detect if something broke)
* there are some maintainability issues in the code (lack of encapsulation, abuse of conditionals, repeated code, repeated strings that are not defined as constants and non unit-testable code)
* lack of unit tests
* Spec is lacking important aspect of access management that would soon need to be covered and might lead to hackish solution based on current implementation (for example, permissions over a specific instance or record, such as you can edit all users except for admin, or a user can't edit other users but he can edit himself, etc.)
* a lot of code it's being written and will have to be maintained, but the code it's not specific to OpenCATS but generic Access control management that we could reuse from open source (such as Symfony)

Is there any chance we could discuss this altogether? Should I ignore all my concerns and just comment on the code specifics?

skrchnavy:
Hi @mlespiau.

We have tests done for current codebase to check existing access levels (https://github.com/AnritsuSolutionsSK/OpenCATS/tree/access_level_tests). This branch is based on branch waiting as PR #100. That branch passed on travis, see https://travis-ci.org/AnritsuSolutionsSK/OpenCATS/branches

then there is this PR with reimplementation of accesslevel checks and we have additional branch with tests for this PR in branch https://github.com/AnritsuSolutionsSK/OpenCATS/tree/security_improvements_tests (currently unfortunately failing o travis)

skrchnavy:
I created PR for access level test (#125) to provide better overview of next steps

skrchnavy:
Regarding of your concern to constants, we discussed with @Kixy25 if constants are useful her but decided that strings are more explaining currently.
for example access level for protection of adding candidates is written in form: `candidates.add`, that mens module `candidates` and action `add`. 
so it means whom to display url `index.php?m=candidates&a=add' and to protect calls to CandidatesUI.handle in switch `"add"` ( current code reference https://github.com/opencats/OpenCATS/blob/master/modules/candidates/CandidatesUI.php#L91 https://github.com/opencats/OpenCATS/blob/master/modules/candidates/Add.tpl#L43 ...)
 Using any constant could be less readable.

Due to better readability were checks from private methods in particular UI files moved to handle method next to action _constant_.
https://github.com/opencats/OpenCATS/pull/91/files#diff-7cc47e2ebae9ea85b8d9be1912011d73R96 vs. https://github.com/opencats/OpenCATS/pull/91/files#diff-7cc47e2ebae9ea85b8d9be1912011d73L914 and https://github.com/opencats/OpenCATS/pull/91/files#diff-7cc47e2ebae9ea85b8d9be1912011d73L941.

skrchnavy:
I checked Symphony and support for requested functionality. In ACL sections they advised for simpler use case (as is ours) to use [Voters](http://symfony.com/doc/current/security/voters.html).
I checked usage and we can later change easily this private ACL to voters replacing 
```php
                if ($this->getUserAccessLevel('settings.addUser') < ACCESS_LEVEL_DEMO)
                {
                    ....
                }
```
with (or similar)
```php
$this->denyAccessUnlessGranted('addUser', $user);
```
and to implement proper voter in the module.

mlespiau:
Hi @skrchnavy can you please resolve the conflicts? I'll do a final review and let's get this merged!

mlespiau:
approved and merged

PR #100, opened 2016-08-03
Tests using behat isolated in test directory

skrchnavy:
Structure of project changed to be able to isolate tests related files and data into extra directory (`test`).
Updated are:
- behat related files including composer
- data and configuration for tests from docker to test directory
- docker file and docker-composer
- travis configuration


skrchnavy:
@mlespiau please check this PR, we plan to add further tests and use this structure and also to add there some scripts to be able to run tests easily also on local computer.

mlespiau:
Hi, 

Compose should not be moved from the base directory, because it's planned to be used to replace all 3rd party dependencies (to simplify it's management). 

What's the motivation for moving it to the test folder?

On `[another branch](https://github.com/mlespiau/OpenCATS/tree/replace-deprecated-simpletest-with-phpunit-and-behat)` I'm working on specific code to get rid of simpletest (legacy unit tests project) by replacing all tests with phpunit and behat tests.

docker/docker-compose-test.yml
```diff
@@ -1,39 +1,44 @@
 version: '2'
 services:
     opencats:
+      container_name: opencats_test_web
```
mlespiau: Why add the container_name directive? Doesn't it make it more verbose? Might we just update the services name?

docker/docker-compose-test.yml
```diff
@@ -1,39 +1,44 @@
...
       image: prooph/nginx:www
       ports:
           - "80:80"
           - "443:443"
       volumes_from:
-        - dataphp
+        - opencatsdata
```
mlespiau: Like this change :)

docker/docker-compose-test.yml
```diff
@@ -1,39 +1,44 @@
...
       volumes:
         - ..:/var/www/public
-        - ./test/fs/config.php:/var/www/public/config.php
-        - ./test/fs/INSTALL_BLOCK:/var/www/public/INSTALL_BLOCK
+        - ../test/config.php:/var/www/public/config.php
+        - ../test/INSTALL_BLOCK:/var/www/public/INSTALL_BLOCK
       command: "true"
 
-    mariadb:
+    opencatsdb:
```
mlespiau: Like this change!

docker/docker-compose-test.yml
```diff
@@ -1,39 +1,44 @@
...
+      container_name: opencats_test_mariadb
       image: mariadb
       ports:
         - 3306:3306
       environment:
-        - MYSQL_ROOT_PASSWORD=dev
-        - MYSQL_USER=dev
-        - MYSQL_PASSWORD=dev
-        - MYSQL_DATABASE=cats
```
mlespiau: Why change the styles here? Is the dash a bad practice? I think it kinds of make it more readable.

docker/php/Dockerfile
```diff
@@ -22,7 +22,8 @@ RUN cd ~ && \
   make install
 RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ && docker-php-ext-install -j5 mysql gd ldap soap zip
 ADD scripts/install-composer.sh /opt/install-composer.sh
-RUN ["chmod", "+x", "/opt/install-composer.sh"] && /opt/install-composer.sh && mv /var/www/html/composer.phar /usr/local/bin/composer
+RUN ["chmod", "+x", "/opt/install-composer.sh"]
+RUN /opt/install-composer.sh && mv /var/www/html/composer.phar /usr/local/bin/composer
```
mlespiau: Separating the commands makes the image size bigger.

Also, it seems the permissions problem that made you add the chmod can be avoided by adding the USER directive (and using ADD or COPY, can't recall the right one now).

docker/docker-compose-test.yml
skrchnavy: This is just display name and is not related to service name. When you see containers, you know, what compose you have installed. Otherwise you see names generated from service name.
Please compare with docker-compose-test.

Service names are in conflict in dev and test dockers, I haven't solve this. Not sure if it is necessary.

docker/docker-compose-test.yml
skrchnavy: I don't know shy are dashes not working for me.  I will recheck if it is reproducible or the cause was somewhere else

docker/php/Dockerfile
skrchnavy: I tested common command and only first command on line was processed and the rest was ignored. In separated commands, all was working. Other option is to remove chmod and run the command as /bin/sh argument. I will test.

skrchnavy:
I see 'test' dir as extra project on top of OpenCATS, so it could be even in another repository. This separation allows in the future to create installation only with necessary files for production ignoring test directory.

So moving compose file to test and not using common in root, I wanted to separate test specific dependencies from production dependencies. Currently I see only test specific dependencies in compose.yml. We can revert root compose file to root directory with no entries to prepare for further dependencies management (I agree this is good approach)

skrchnavy:
I checked how composer works and agree with you that it shall be in root dir. useful information is here: https://seld.be/notes/composer-installing-require-dev-by-default to explain differencies between dev and prod scopes.

skrchnavy:
Strucure was updated based on comments from @mlespiau , please review

docker/docker-compose-test.yml
skrchnavy: reverted to use dashes

docker/php/Dockerfile
skrchnavy: RUN in form ```RUN [ "cmd", "argument", ...]``` can't be followed by next commands even with && so have to be split to 2 run commands. so it was corrected in 2ab6f9541e329932e7c2833ddbf664510e7c9c42

RussH:
@mlespiau @skrchnavy is this reviewed and tested? As it's all behat.. had better rely on you to advise!

skrchnavy:
I tested it and it works OK. @mlespiau started his update of test cases, so he shall approve it not to break his changes. It requires few files to be moved to new place and some updates in his travis.yml file
We are building our tests using this branch as baseline.

RussH:
Okay thanks will wait for @mlespiau !



----
Russ

mlespiau:
Hi @skrchnavy. I'd been thinking deeply about moving the test code and reviewing what other projects are doing.

The main issues I have with this change are the following: 
* I believe it goes agains the namespaces / bundles best practices
* Separating the tests from the code it's not really required, what we really need is to separate customer data from code   

**Namespace**
One of the purposes of Namespacing and PSR-4 is to allow meshing up/grouping code from different projects together while at the same time avoid naming clashes that might exist between two projects or bundles.

Although not a standard, many dependencies that are managed through composer and also [Symfony bundles](http://symfony.com/doc/current/bundles/best_practices.html) include a Test folder inside the namespace with their test code.

I think we could follow the same approach, maybe it would even come handy to allow the project to be installed through composer.

**Should we separate test code from production code? What's separation that matters?**
I've checked and searched in google but could not find any references about deploying test code to production being a bad practice or either not deploying the test code to production being a good practice. 

What's the value that this could provide to you? Why would you like to achieve this?

Nevertheless, there is a separation that's really important, and that I mentioned in the [OpenCATS forums](https://groups.google.com/forum/#!topic/opencats-dev/53EB-U8pdIc). Separating data from code. Attachments, session files, uploads, configuration files, etc. should be separated from the code to simplify deployments. 

If you agree with my comments above, there is also a change that must be done, and it's to move the behat tests to src/OpenCATS/Tests/FunctionalTests folder. 

At last, I don't want to reject this pull request if it really yields value to you but would like to get more context.

skrchnavy:
@mlespiau, this pull request was motivated to remove dependency of tests on Docker, so that tests and test data are outside docker structure and Docker is partially dependent on the test (which is OK because there is docker-compose-test).
This isolation simplifies running tests in local WAMP installation during development and have all tests related artefacts together. (It would be nice also to run tests on travis without docker but I was not successful in implementation)

These changes are in this branch
- test.sql moved from docker to test dir
 - docker-compose-test updated accordingly
- composer.json updated to have mink as dev-dependency
 - I am not sure if it is planed to be used in production (could be reverted if this is not good).
- behat.yml and feature directory moved to test subdir
 - this is on discussion, could be reverted if you see it not good way. Or later to move it where you proposed.
- behat.yml updated to use suites and contexts to be extensible
 - we use another test data for security tests then are used in original tests
 - I think features shall be in the future separated into subdirs by suites 
- docker-compose files updated to name containers
 - when I install particular setup I can see, if I have test docker environment or production environment in docker deployed.

So I think above changes are align with your goal to separate data from code (for example test.sql is not production data nor code).
As I stated test code (features+behat.yml) can be reverted back to original structure. But then there will be lot of dirs and files not related to production so install scripts could be more complex. But compose may have cheap solution for this.

docker/docker-compose.yml
```diff
@@ -1,45 +1,49 @@
...
         - 3306:3306
       environment:
-        - MYSQL_ROOT_PASSWORD=dev
+        - MYSQL_ROOT_PASSWORD=root
         - MYSQL_USER=dev
         - MYSQL_PASSWORD=dev
         - MYSQL_DATABASE=cats
       volumes:
-        - ./data:/docker-entrypoint-initdb.d
+        - ../test/data:/docker-entrypoint-initdb.d
```
skrchnavy: @mlespiau I propose to remove this init of DB in docker-compose (production variant) or link it to production (empty) database if necessary

PR #117, opened 2016-08-18
updated styling for login screen

RussH:


RussH:
screenshot_20160818_184430

RussH:
just need to test locally;
- can I wipe install block and reinstall without weirdness.

PR #122, opened 2016-08-25
fixed send email and add candidates

### LINKED ISSUE #121: No actions work against list members in 0.9.3
reproducable on 0.9.3 at http://demo.opencats.org;

when a list of candidates is created, all actions fail, including 'send email' and 'add to pipeline'. 

all 'Actions' fail on candidates to - seems to be the Actions feature that's failing everywhere.

skrchnavy:
fixed #121

### DIFF
--- a/modules/candidates/CandidatesUI.php
+++ b/modules/candidates/CandidatesUI.php
@@ -1327,11 +1327,6 @@ class CandidatesUI extends UserInterface
             CommonErrors::fatal(COMMONERROR_PERMISSION, $this, 'Invalid user level for action.');
         }
         
-        if (!$this->isRequiredIDValid('candidateID', $_GET))
-        {
-            CommonErrors::fatal(COMMONERROR_BADINDEX, $this, 'Invalid candidate ID.');
-        }
-        
         /* Get list of candidates. */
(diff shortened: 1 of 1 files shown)


RussH:
thanks!

mlespiau:
Hi, what's the root cause of this issue? Why are this validations not required any more?
Shouldn't we cover this fix with tests?

skrchnavy:
Hi. These checks were not there before when it worked. I agree there shall be test but with tests we implemented this was not caught.
Check in https://github.com/opencats/OpenCATS/commit/00166b4c3c83d0dcc9b7627411b144cb0d77758c#diff-7cc47e2ebae9ea85b8d9be1912011d73L1330 was probably too strict, there are further conditions to use different sources of candidateIDArray.

Check in https://github.com/opencats/OpenCATS/commit/00166b4c3c83d0dcc9b7627411b144cb0d77758c#diff-7cc47e2ebae9ea85b8d9be1912011d73L3244 is strange, because there is call to method in https://github.com/opencats/OpenCATS/blob/0c11685fcd4b8ce258817722c0144a6bb4bc07d9/lib/DataGrid.php#L293 where is same check that didn't work. `if (!isset($_REQUEST['i']) || !isset($_REQUEST['p']))` but is called trigger_error function.

PR #124, opened 2016-08-26
Update InstallationTests.php

theSlyest:
I updated the **checkSOAPExtension()** method since it was always returning _false_ on newer versions of PHP. I replaced **is_callable()** by **method_exists()**. It seems that **is_callable()** returns _true_ only for static methods when used with a class name.

### DIFF
--- a/lib/InstallationTests.php
+++ b/lib/InstallationTests.php
@@ -342,9 +342,10 @@ class InstallationTests
     public static function checkSOAPExtension()
     {
         /* Is the SOAP extension loaded?. */
+        /* The is_callable function seems to work with   */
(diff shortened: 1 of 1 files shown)


mlespiau:
Hi, I believe we should create some tests for this use case. It can be tricky but we could at least cover the logic by extracting the parts of the conditional into its own class and then mocking the method to test how the application behaves when the soap library is either installed or not installed, etc.

RussH:
@theSlyest did you want to update this following comments from @mlespiau ?

theSlyest:
I proposed this patch because of the wrong result this method was returning in my own server. But I think it would be better to confirm the problem - and even the solution - by performing other tests.

PR #132, opened 2016-09-22
Upgraded ckeditor to current (version 4.5.11) to address #131

### LINKED ISSUE #131: While sending email to candidates the body field does not appear
Hi 
When I select a (one or a set of) candidate(s) and try to send them email this is what appears

image

is there some way the body field also start appearing???

RussH:
Upgraded ckeditor to current (version 4.5.11) to resolve  #131


RussH:
tested at http://demo.opencats.org

PR #158, opened 2016-11-14
Extended xml export

skrchnavy:
Added notes and type into xml export 
also fixef few issues in xml export


RussH:
I can't see any issue with this, I assume it doesn't impact your larger PR's that are waiting?

RussH:
Enhancement.. Accepted!

skrchnavy:
Thanks for comments, we will extend this export also with one additional parameter.
We were not sure how it impacts Rss channel but we found special Rss class for that, so hope no influence on standard functionality.

PR #159, opened 2016-11-15
Configurable Job Types 

skrchnavy:
Motivation
---

JobTypes were hardcoded and spread across code, only these job types were defined: 'Contract', 'Hire', 'Contract to Hire', 'Freelance'. To add new contract type, iw would require to add code for new type into few functions.

Proposed solution
---

This PR allows to extend/change list of jobTypes in configuration file (config.php).

(moving this list to DB would require changes in DB and also in GUI to be able to extend it. For future it could be useful).


lib/JobOrders.php
```diff
@@ -809,28 +809,11 @@ public function updateOpeningsAvailable($jobOrderID, $count)
      */
     public static function typeCodeToString($typeCode)
     {
-        switch ($typeCode)
+        if(!defined('JOB_TYPES_LIST') || JOB_TYPES_LIST[$typeCode] == null)
```
skrchnavy: @Kixy25 please move this check to extra function that would return custom job types list or default job type list that is 100% compatible with current 'hard-coded' list.

modules/joborders/JobOrdersUI.php
```diff
@@ -591,6 +591,14 @@ private function add()
         $careerPortalSettings = new CareerPortalSettings($this->_siteID);
         $careerPortalSettingsRS = $careerPortalSettings->getAll();
         $careerPortalEnabled = intval($careerPortalSettingsRS['enabled']) ? true : false;
+        
+        if(defined('JOB_TYPES_LIST'))
```
skrchnavy: this shall be solved also using a function in ib/JobOrders.php

modules/joborders/JobOrdersUI.php
```diff
@@ -863,7 +872,14 @@ private function edit()
                 }
             }
         }
-
+        
+        if(defined('JOB_TYPES_LIST'))
```
skrchnavy: this shall be solved also using a function in ib/JobOrders.php

RussH:
This works for me - I have amended the contract types in my installation by hand.. Having this as a selection is far easier. Waiting for comments from @mlespiau  / @matiaslespiau

config.php
```diff
@@ -276,4 +276,11 @@
 
 define ('LDAP_SITEID', 1);
 
+const JOB_TYPES_LIST = array(
```
skrchnavy: This shall be defined as example in config.php with short description how to use it.

mlespiau:
I really like the change. It's heading on the right direction. 

Actually we should make the JobOrderTypes configurable on the application (and store them in the DB). But this change is a simpler intermediate step. 

The implementation seems clean. I would suggest creating a new class JobOrderTypes instead of placing the methods for getting the JobOrderType names/list on the JobOrder class.

libregeek:
I have this implemented using DB in my private fork. I can contribute the code if you are interested. The only downside is that it might take couple of days as I'm overloaded with official work. Please let me know your thoughts.

skrchnavy:
In current PR, there is class for JobOrderTypes, so this can be extended later to use JobOrdersTypeEntity (based on @mlespiau introduced concept).
@libregeek please if it is possible to publish it on github.

lib/JobOrderTypes.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+/**
+ * CATS
```
skrchnavy: this header shall be changed to GPL, @RussH do we have header example?

RussH:
Lets use this;
`<opencats candidate applicant tracking system.>
Copyright (C) <2017>  <name of author>

This file is part of opencats.

opencats is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
the Free Software Foundation, either version 2 of the License, or
(at your option) any later version.

opencats is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU General Public License for more details.

You should have received a copy of the GNU General Public License
along with opencats.  If not, see <http://www.gnu.org/licenses/>.`

skrchnavy:
Are there any further comments?

mlespiau:
Approved, great job!

PR #167, opened 2016-11-28
Zipcode address completion via google

### LINKED ISSUE #38: Canadian Postal Codes
Is there value in adding canadian postal codes to the zipcode table?

I have the full list.

RussH:
This is a working update to allow candidate address completion by a zipcode lookup using a google api. The api allows free usage of 2,500 queries per day per IP address. Of course, google changes it's api's frequently - but if we use this we can remove the 'zipcode table' from the database.

This closes  #38 

I will have to add some comments into documentation as different countries may want to return different values from the lookup for City and State - see the comments on #38.

Coding entirely provided by Dale Stringfellow ( @thedss ), with testing / feedback  from me

### DIFF
--- a/ajax/zipLookup.php
+++ b/ajax/zipLookup.php
@@ -1,30 +1,7 @@
 <?php
 /*
- * CATS
- * AJAX City/State lookup via Zip Interface
- *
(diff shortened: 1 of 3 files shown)


RussH:
The code for this is already deployed to demo.opencats.org and works successfully there

PR #168, opened 2016-11-29
Module for handling of duplicate candidates

skrchnavy:
New module for handling of duplicate candidates added and integrated into other modules (candidates).
Workflow:
- When candidate added, duplicity is checked and link between candidates is created
- Linked candidates are visible in candidates (for proper access level)
- Duplicates module is used to merge duplicates or remove link between duplicates


skrchnavy:
Main functionality implemented, missing tests (functional and security) and access level checks in module/duplicates.
Automatic merge not planed to prevent merging of not real duplicates. But such improvements based on configuration are possible later.

@mlespiau  / @RussH  please review / comment the feature.

libregeek:
Why is the duplicate checking not part of the Candidate module? AFAIK, it's using only the candidate data and not related to any other modules. I would suggest we should go for a new module only if it has a new entity, otherwise it's just making things too complex.

skrchnavy:
The reason is to be able to allow such functionality only for selected users and to have it separated.
But if it is preference to add it as a part of Candidate, we can move it there as submenu(s).

lib/TemplateUtility.php
```diff
@@ -1123,14 +1123,23 @@ public static function filterRemoveTextBlock($text, $startBlock, $endBlock, $clo
...
+        {
+            echo '<a href="javascript:void(0);" onclick="showHideSingleQuickActionMenu('.$dataItemType.', '.$dataItemID.', docjslib_getRealLeft(this)-20, docjslib_getRealTop(this)+20);"><img src="images/downward.gif" border=0></a>';
+        }
+        else
+        {
+            $url1 = "'".$url1."'"; 
+            $url2 = "'".$url2."'"; 
+            echo '<a href="javascript:void(0);" onclick="showHideSingleQuickActionMenuExtended('.$dataItemType.', '.$dataItemID.', docjslib_getRealLeft(this)-20, docjslib_getRealTop(this)+20, '.$url1.', '.$url2.');"><img src="images/downward.gif" border=0></a>';
+        }
     }
```
mlespiau: Again, adding conditionals due to lack of abstractions as explained in quickActionMenu js code. Check PR.

skrchnavy:
Duplicates are now part of candidates module

config.php
```diff
@@ -433,6 +433,12 @@
             'candidates.emailCandidates'
             'candidates.show_questionnaire'
             'candidates.list'
+            'canidates.viewDuplicates'
+            'canidates.linkDuplicate'
+            'canidates.merge'
+            'canidates.mergeInfo'
+            'canidates.removeDuplicity'
+            'canidates.addDuplicates'
```
mlespiau: There is a small typo "canidates" instead of "candidates" in all the new entries.

constants.php
```diff
@@ -60,6 +60,7 @@
 define('DATA_ITEM_USER',        600);
 define('DATA_ITEM_LIST',        700);
 define('DATA_ITEM_PIPELINE',    800);
+define('DATA_ITEM_DUPLICATE',   900);
```
skrchnavy: this is used in quickActionMenu (file module/candidates/ShowDuplicates.tpl)

RussH:
@mlespiau  regarding handling schema changes during upgrades..

this seems to be handled here; https://github.com/opencats/OpenCATS/tree/master/modules/install/scripts

as there have been schema changes in the past and this seems to accommodate changes requried during upgrades.

constants.php
mlespiau: ok!

skrchnavy:
Update is handled in this code file: https://github.com/opencats/OpenCATS/blob/master/modules/install/Schema.php
update script have to be added into https://github.com/opencats/OpenCATS/tree/master/modules/install/scripts and also into code.

Do we have documentation how update works and how it shall be handled from release to release?

RussH:
@skrchnavy you asked about the handling of updates. No there's no documentation regarding the handling of updates - beyond the areas you've identified in 

Code updates;
https://github.com/opencats/OpenCATS/blob/master/modules/install/Schema.php
update script;
https://github.com/opencats/OpenCATS/tree/master/modules/install/scripts and also into code.

skrchnavy:
Based on comments from our recruter, duplicate management will be cahnged
- view candidate with duplicate will be merged with standard candidate view
- candidates list will be merged with duplicates list (no extra submenu for duplicates)
- added will be quick filter button to select only candidates with duplicates (similar to hot/my candidates).

for now it is not planed to include (due to lack of time)
- management of duplicate identification (which fields to be used)
- search duplicates form.

modules/install/ajax/ui.php
```diff
@@ -874,10 +874,15 @@
             /* 0.6.0 */
             $revision = 60;
         }
-        else if (isset($tables['history']))
+        else if (!isset($tables['candidate_duplicates']))
```
skrchnavy: @mlespiau 
>How do we migrate to the new schema? We need some sort of tool that applies the SQL update. What do we as for now? Moving forward if you agree in moving to Symfony we can use it's data migrations functionality.

this code checks presence of tables and then calls upgrade script.
Currently db/upgrade-0.9.4-0.9.5.sql because it is expected to be part of 0.9.5 version. When plan changed, this part shall be corrected accordingly and also file with create table renamed._

_migration was not tested yet._

RussH:
I manually tested this against the develop branch. All works, but during installation had to manually run the mysql upgrade against the database (admittedly tested against a v0.9.1a backup). @skrchnavy What's the method for calling the upgrade script?

skrchnavy:
In this PR, there is an upgrade script: db/upgrade-0.9.4-0.9.5.sql
And there is also check for version in ui.php to run this script.

Due to release of 0.9.4, this have to be fixed to call it properly.

RussH:
Great, so we can roll this in and then fix the script to call it afterwards?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

skrchnavy:
This shall identify possible duplicates if I remember. I need to check.

PR #172, opened 2016-12-01
Deprecated: Function split() is deprecated

Bloafer:
Deprecated: Function split() is deprecated on PHP 5.5.9


Bloafer:
Checks failed on Travis due to "Error Connecting to Database" not because of PR

Issue #94, opened 2016-07-26
Issue with showing Profile Picture when changing


ManuelNL:
If you want to change/edit the candidates profile picture in OpenCats it didn't show the current picture. Just a empty square.
The changes you have to make are minor.

You have to edit the file: ...\modules\candidates\CreateImageAttachmentModal.tpl

At line 12 and 13 you wil find:

<a href="attachments/<?php $this->_($attachmentsData['attachmentID']) ?>/<?php $this->_($attachmentsData['storedFilename']) ?>">
<img src="attachments/<?php $this->_($attachmentsData['attachmentID']) ?>/<?php $this->_($attachmentsData['storedFilename']) ?>" border="0" width="165">
</a>

Change : attachmentID into directoryName

It will look like this

<a href="attachments/<?php $this->_($attachmentsData['directoryName']) ?>/<?php $this->_($attachmentsData['storedFilename']) ?>">
<img src="attachments/<?php $this->_($attachmentsData['directoryName']) ?>/<?php $this->_($attachmentsData['storedFilename']) ?>" border="0" width="165">
</a>

The allready changed file:
[CreateImageAttachmentModal.zip](https://github.com/opencats/OpenCATS/files/384000/CreateImageAttachmentModal.zip)

RussH:
Manuel's comments form the forum.. 
If you want to change/edit the candidates profile picture in OpenCats it didn't show the current picture. Just a empty square.
The changes you have to make are minor.

You have to edit the file: ...\modules\candidates\CreateImageAttachmentModal.tpl

At line 12 and 13 you wil find:

```
<a href="attachments/<?php $this->_($attachmentsData['attachmentID']) ?>/<?php $this->_($attachmentsData['storedFilename']) ?>">
<img src="attachments/<?php $this->_($attachmentsData['attachmentID']) ?>/<?php $this->_($attachmentsData['storedFilename']) ?>" border="0" width="165">
</a>
```

Change : attachmentID into directoryName

It will look like this

```
<a href="attachments/<?php $this->_($attachmentsData['directoryName']) ?>/<?php $this->_($attachmentsData['storedFilename']) ?>">
<img src="attachments/<?php $this->_($attachmentsData['directoryName']) ?>/<?php $this->_($attachmentsData['storedFilename']) ?>" border="0" width="165">
</a>
```

Save the file and the problem is solved.

Happy recruiting

Issue #4, opened 2014-10-10
When changing picture candidate, it doesn't show current picture


ManuelNL:
You can easily add a picture to a candidate.
If you want to change it (nobody does I think) it doesn't show the current picture.
The link to the picture is wrong.

RussH:
Fixed by PR #173

PR #173, opened 2016-12-02
A patch to show candidate profile picture when it's being changed.

RussH:
Closes Issue #94  - patch from @ManuelNL


RussH:
Also closes #4

PR #181, opened 2016-12-07
Used mysqli procedural methods instead of mysql methods.

### LINKED ISSUE #171: Installation failed in PHP-7
Tried to install OpenCATS (master branch) in PHP-7.0.13 (Fedora 25 running MariaDB-10.1.19).  Following is the error message:

> PHP MySQL extension (mysql) is not loaded.
> Check your settings in php.ini.

The above error is due to the following checking in index.php
```
if (!function_exists('mysql_connect') || !function_exists('session_start'))
{
    die('CATS Error: All required PHP extensions are not loaded.');
}
```
and 
if (!self::DEBUG_FAIL && extension_loaded('mysql') && function_exists('mysql_connect'))
        {
            echo '<tr class="pass"><td>PHP MySQL extension (mysql) is loaded.</td></tr>';
            return true;
        }
mysql_connect is deprecated as of PHP-5.5 and it's removed in PHP-7.
Refer: http://www.php.net/mysql_connect

I think we should use extension_loaded() method instead of function_exists(). 
http://php.net/manual/en/function.extension-loaded.php

I think we should change mysql_connect to mysqli_connect. 

Do we need to maintain backward compatibility till PHP-4?

libregeek:
Fixes #171 

This is a quick fix to make OpenCATS compatible with PHP7. Please note that this PR is just for mysql and I haven't touched any other methods deprecated in PHP-7. This is tested only in PHP-7.0 and MariaDB-10. I didn't execute any automated test cases. I hope it will be take care in travis.

There are several things to be improved:
1. Use Object Oriented methods for mysqli or move the entire DB stuff to PDO.
2. The DB connection is instantiated in multiple files. This needs to be unified to use a single DB object. This will also make it easy to fix the binding methods.

### DIFF
--- a/docker/php/Dockerfile
+++ b/docker/php/Dockerfile
@@ -20,7 +20,7 @@ RUN cd ~ && \
   ./configure && \
   make && \
   make install
-RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ && docker-php-ext-install -j5 mysql gd ldap soap zip
+RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ && docker-php-ext-install -j5 mysqli gd ldap soap zip
 ADD scripts/install-composer.sh /opt/install-composer.sh
 RUN dos2unix /opt/install-composer.sh && \
   chmod +x /opt/install-composer.sh && \
--- a/index.php
+++ b/index.php
@@ -124,9 +124,9 @@ if (ini_get('session.auto_start') !== '0' &&
 }
 
 /* Proper extensions loaded?! */
-if (!function_exists('mysql_connect') || !function_exists('session_start'))
+if (!function_exists('mysqli_connect') || !function_exists('session_start'))
 {
-    die('CATS Error: All required PHP extensions are not loaded.');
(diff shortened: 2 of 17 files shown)


skrchnavy:
Travis tests failed with  this error
```OpenCATS Error: Either PHP Sessions extension or MySQLi extension is not loaded.```

mlespiau:
In order to install it in docker you can just change the following line in `docker/php/Dockerfile`: 

`RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ && docker-php-ext-install -j5 mysql gd ldap soap zip opcache`
with
`RUN docker-php-ext-configure gd --with-freetype-dir=/usr/include/ --with-jpeg-dir=/usr/include/ && docker-php-ext-install -j5 mysqli gd ldap soap zip opcache`

mlespiau:
hi @libregeek, any plans to continue working on this PR?

libregeek:
@mlespiau I guess everything worked good in my development environment. But I may have to revisit it.

libregeek:
Sorry, I didn't realize that it was an old branch. Please let me know if I need to do a fresh pull request after.

RussH:
Will look to apply the PHP v7 compatibility changes after all other PR's

PR #182, opened 2016-12-07
Removed check of job order limits

skrchnavy:
Code cleanup to remove part of code that is not used.
There is no set of `overOpenJOQuota` nor `FREE_ACCOUNT_JOBORDERS`


RussH:
Thanks Sveto - this is simple cleanup, so as long as it builds okay, I'll merge it without waiting for Matias!

PR #183, opened 2016-12-08
Added missing speech mark

Bloafer:
which caused PCLS to not work


Bloafer:
Failed because of build environment not code change

RussH:
typo amendment, doesn't require waiting for Travis to be fixed.

PR #199, opened 2017-01-06
Fixes Non-static method

Bloafer:
Strict standards: Non-static method FileUtility::makeSafeFilename()


skrchnavy:
this function is used only in one place and as static, so correct fix.

mlespiau:
Thanks @Bloafer!

PR #201, opened 2017-01-09
Fix of access level check in Edit user

skrchnavy:
modules settings/EditUser.tpl, conversion from accessLevel to getAccessLevel


mlespiau:
Looks good.

PR #203, opened 2017-01-11
CKEditor replaces tynyMCE

### LINKED ISSUE #202: different HTML WYSIWYG editors used
In the code there are used
*  [tinyMCE](https://www.tinymce.com/) as primary editor for job orders
* [CKEditor](http://ckeditor.com/) as eitor for HTML emails

this shall be unified to use only one.
What is preferred choice?

We have one requirement to have possibility to setup fonts including size.

skrchnavy:
Resolved #202 
CKEditor is used as HTML WYSIWYG editor also in editing of job orders.
Also font plugin in CK editor added.


mlespiau:
Tested it! Works fine.

PR #205, opened 2017-01-11
Pipeline grid contains reference number

skrchnavy:
Job Order could have not unique name so it is necessary to display also reference number of job order in candidate view.


skrchnavy:
I agree with the comment but when you look at the history these files are changed after 5 years. I suggest to wait with optimisation to new templating system and ORM.

mlespiau:
Ok. As it's not a major issue, I approve and merge.

PR #209, opened 2017-01-16
Filled positions cannot hire

skrchnavy:
To allow to add candidate only into job orders that are allowed to accept candidates.


mlespiau:
The design could be improved - this feature is adding if-cases in at least two places of the code when we could use polymorphism instead for a cleaner design. 

I'll approve as I believe we can fix these design issues once symfony becomes the new standard.

PR #225, opened 2017-03-16
TinyMCE – r has no properties error.

christallization:
On line: 4846 I replaced: if(r.item)return false; with if(r==null||r.item)return false;
Recreate on Add Job Order page: /index.php?m=joborders&a=add&jobOrderID=11&typeOfAdd=new
Occurs in Chrome and Firefox and logs to console when there’s a textarea in a hidden div.


skrchnavy:
In PR #203 is removed tinyMCE from the tool, CKEditor is used.
Anyway, tinyMCE is/was not developped in this project, bugs shall be reported into https://github.com/tinymce/tinymce.

RussH:
Hi @christallization as @skrchnavy has replied above, in the development version we have completely removed TinyMCE, replaced with CKEditor throughout. Therefore I'm going to close this patch request. We do very much thank you for the effort though, and we're aiming to release the current code as soon as possible. Please continue to hack on the codebase :-D

PR #230, opened 2017-03-30
Job order import

skrchnavy:
Extending import functionality with import of job orders


modules/import/Import.php
```diff
@@ -437,6 +437,122 @@ public function add($dataNamed, $userID, $importID)
...
+
+    /**
+     * Adds a record to the joborder table.
+     *
+     * @param array (field => value)
+     * @param userID
+     * @param importID
+     * @return joborderID
+     */
+    public function add($dataNamed, $userID, $importID)
```
skrchnavy: `import_id` field is missing in database in table `joborder` so not used.

RussH:
Okay, so I tested this with the attached sample joborders 
[sample-job-import.txt](https://github.com/opencats/OpenCATS/files/1098425/sample-job-import.txt)

the import says successful, 4 joborders imported - but there are none listed in the joborder tab. 

Then, when I return to settings/administration/data import / view recent imports it will tell me
image

If I try to revert the (possibly worked?) import I get a query error;
image

RussH:
Built and tested locally - all works fine

Issue #127, opened 2016-09-18
Subject Modification in email Alert Templates + Add New Email Alerts Template


KamalJoshiDelhi:
Hi,
I am using OpenCATS-0.9.3-1 and was looking to modify the Subjects in the available Templates and also ability to add new Templates.

Regards
Kamal Joshi

RussH:
multiple email templates - This is/was a feature documented here, but never introduced to the main codebase. 
[CATS-20130807-v0.2.pdf](https://github.com/opencats/OpenCATS/files/606885/CATS-20130807-v0.2.pdf)

RussH:
Planned in #231

skrchnavy:
In #231 there is implemented edit template feature but it is not possible to choose email template for status changes as is described in document. So #231 partially implements this.

RussH:
Making email template subjects editable and allowing additional templates would be useful, but it isn't currently on the OpenCATS roadmap and isn't something we're likely to deliver in the near future.

I've added the request to FEATURE_REQUESTS.md so it remains recorded. Closing this for now rather than leaving it as an active issue.

PR #231, opened 2017-03-30
Email templates support

skrchnavy:
User is able to define own templates and delete user defined templates (the original ones cannot be deleted).
User is able to use them for direct emailing to candidates e.g. from main candidate datagrid under `Action ->Send Email`.


skrchnavy:
Partially related to #127

RussH:
@skrchnavy this tests out fine on my local test box - but an enhancement request before I merge.. after sending candidates emails is there any way to have the activity list for the recipients updated with a copy of the emails sent? i.e. 'sent email' <email body>

PR #236, opened 2017-04-06
Job order statuses configurable

skrchnavy:
Added support for configurable job order statuses
Default implementation contains statuses as were until now but user can extend (or change) statuses in config.php. Statuses are also grouped into stages (currently 3 stages/groups - Pre-Open, Open, Closed).
Further configuration is possible for filters in job order page, 'shared' pages (xml, rss, carrer portal),  and statistics.


RussH:
@skrchnavy  - does this break the wordpress plugin?

skrchnavy:
@RussH I am not sure but looking at the code, I see there 2 places with select from joborder where is status. This change is compatible if 'active' status used for original purpose.
I think the plugin have one small issue related to collation, it expects DB is case insensitive, because it checks `active` status and in DB there is `Active` status (you can see it used as default value in schema)

db/cats_schema.sql
```diff
@@ -792,7 +792,7 @@ CREATE TABLE `joborder` (
   `duration` varchar(64) COLLATE utf8_unicode_ci DEFAULT NULL,
   `rate_max` varchar(255) COLLATE utf8_unicode_ci DEFAULT NULL,
   `salary` varchar(64) COLLATE utf8_unicode_ci DEFAULT NULL,
-  `status` varchar(16) COLLATE utf8_unicode_ci NOT NULL DEFAULT 'Active',
+  `status` varchar(64) COLLATE utf8_unicode_ci NOT NULL DEFAULT 'Active',
```
skrchnavy: This is for new installation and in case, when status is longer then 16 chars. Statuses in default configuration are shorter.
We shall add into documentation how to alter DB manually for older installation in case of user defined statuses.

kmharper:
Can anyone provide instructions for implementing this change on an existing site, or is that way too premature? I have a client who would love to have this functionality.

skrchnavy:
Hi @kmharper 

You can clone/fork [AnritsuSolutionsSK/OpenCATS](https://github.com/AnritsuSolutionsSK/OpenCATS) repo and then to use branch [job_order_statuses_configurable](https://github.com/AnritsuSolutionsSK/OpenCATS/tree/job_order_statuses_configurable) as your installation (or [integration](https://github.com/AnritsuSolutionsSK/OpenCATS/tree/integration) branch where are most merge request into main OpenCATS repo merged)

In both branches there are included all changes in master and develop branches.

RussH:
I'm testing locally

geniusupgrader:
How does this work with the Hiring Overview Chart?

RussH:
tests out fine, needs some update into the OpenCATS documentation  to describe custom joborder statuses and joborder types that are configurable in Config.php. Should only need to be a mention. 
@cptr13 do you want me to put some words together?

PR #238, opened 2017-04-11
Import of files in different encoding

skrchnavy:
added option to select encoding in imported file when list of encoding configured in config file.
First encoding in the list is default.
If list not defined, option box is not displayed.


mlespiau:
Hi, 

**I have a few comments regarding this feature:** 
1. Why should our users know which encoding to use? Can't we just simplify it by using http://php.net/manual/en/function.mb-detect-encoding.php to auto-detect the encoding of the uploaded file?
2. Why is the list encoding configurable? I believe it should be fixed and we should support all possible encodings. 

**And some comments regarding this code:** 
1. The modules/import/Import.php file contains too many classes. Each class should be moved to it's own class.
1. The code for each class it's pretty similar and duplicated. The same exact lines need to be added three times, making the code harder to maintain. As the three classes has the same API they could benefit from inheriting from a base abstract class (with a fancy name such as ImportableEntity).  then the similar code could be placed on the base class and reused by it's children. 
1. Checking for $encoding == '' or null it's an anti pattern. A better approach is to use objects instead of strings and use [NullObject design pattern](https://sourcemaking.com/design_patterns/null_object). 

**And a general comment about the design:**
IMHO, this it's not right place to check for encoding and convert as it can lead to multiple bugs as the codebase grows. In general it's a good idea in order not to make mistakes with encoding to:
1. Assume everything is UTF-8
1. When doing a file upload or getting files from external sources, convert to UTF-8 as first action and the pass to the the rest of the code. 
Trying to handle UTF-8 through the code creates more complexity and duplication (as you add feature and maintain the codebase, you need to start checking for encoding and convert in more an more places).

skrchnavy:
Wrt.  _And some comments regarding this code:_
This PR solved encoding and was not intended to do refactoring / optimisation of the code. There are lot of places where such changes are necessary, is it worth to do these changes prior to symfony migration when some functionality will be migrated anyway?

Rest - I will discuss internally your comments and why it was implemented that way.

mlespiau:
It's not for optimization of the code, but to keep it maintainable. 
I don't think Symfony would change any code at all so we should do it before merging the code IMHO.

Kixy25:
I think mb_detect_encoding is not very reliable from what I've found around the internet. Also it probably supports only these encodings http://php.net/manual/en/function.mb-list-encodings.php and some are missing e.g. Windows-1250 (a need for this encoding was the reason why I started this feature at all). 

If we decide that some missing encodings are okay, then it can be implemented the way you suggested. Configurable encodings from the config.php can be used in [mb_detect_encoding()](http://php.net/manual/en/function.mb-detect-encoding.php) as the $encoding_list, else it takes the list from [mb_detect_order()](http://php.net/manual/en/function.mb-detect-order.php) which in my case listed only UTF-8 and ASCII. The other option would be to use the result of [mb_list_encodings()](http://php.net/manual/en/function.mb-list-encodings.php) which should give us all supported encodings and then we can ditch the configurable option from config. I don't know how well that would work because I haven't tested it and the order in the $encoding_list is important so if the mb_list_encodings() doesn't return it in the best possible order it might be useless after all.

As for the division of Import.php, I already implemented that. 
Checking for $encoding=='' is no longer needed because mb_detect_encoding() returns only the encoding or false.
Right now I am implementing the encoding detection  right after fopen and encoding conversion after fgetcsv, not further in the code as before.

Kixy25:
Another option would be to implement import for Excel spreatsheats, I converted my .csv back into .xlsx in Excel and PHP read it just fine with encoding and all (using PHPExcel library). Haven't tried it for import but I don't see why it should not work.

skrchnavy:
@mlespiau if any optimization is expected, it shall be extra PR, mixing any new feature if highway to hell, if anything not working then it could be cause of optimization or new feature and many times difficult to manage.
Optimization shall be done prior to new feature and then on top of it to implement the feature.

mlespiau:
Hi @skrchnavy I'm a strong believer that developers should be refactoring code all the time. According to Martin Fowler's book: https://www.csie.ntu.edu.tw/~r95004/Refactoring_improving_the_design_of_existing_code.pdf

When should you refactor? (p 49) it says basically all the time, but also highlights to refactor when fixing a bug and when doing code reviews.

Now, if your concern is breaking things, that must be addressed as well with good tests. 

Going back to the feature, I'm still not convinced with this functionality. Wearing a "product owner hat" I don't think that Recruiter needs to be exposed to picking the "encoding"  for the file they are uploading. Encoding it's a deep technical concept not related to recruiting at all! @RussH do you have an opinion on this?

Going back to this PR, I have created [a blog post ](https://gignus.ghost.io/using-metrics-to-drive-teams-focus-or-behaviour-change/) for this code review. I'm not done yet. The changes I propose can be seen on this branch:  https://github.com/mlespiau/OpenCATS/tree/import_encoding

mlespiau:
hi @skrchnavy let's not block this PR. I'm approving feel free to merge it. I will provide my suggestions as subsequent PRs.

RussH:
@skrchnavy can you check out why this build is failing please? Suspect it's my resolution of the conflicts. Appreciated.

RussH:
still have to amend code as per https://github.com/opencats/OpenCATS/issues/332#issuecomment-397766063 for career portal to work

PR #245, opened 2017-04-20
Fix state location mapping

### LINKED ISSUE #244: Zipcode lookup populating State field incorrectly
When I enter a zipcode and click the Lookup button, it's putting the county into the State field.  

E.g. using the zip 02210 results in Boston in City (correct) and Suffolk County in State (incorrect).

smackjack3:
Replace administrative_area_level_2 which returns the county with administrative_area_level_1 which returns the state (expected).

resolves #244

### DIFF
--- a/lib/ZipLookup.php
+++ b/lib/ZipLookup.php
@@ -34,7 +34,7 @@ class ZipLookup
 				if (($value->type == 'postal_town') && ($aAddress[2] == '')) {
 					$aAddress[2] = (string) $value->long_name;
 				}
-				if ($value->type[0] == 'administrative_area_level_2') {
+				if ($value->type[0] == 'administrative_area_level_1') {
 					$aAddress[3] = (string) $value->long_name;
 				}
 			}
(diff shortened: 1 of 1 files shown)


RussH:
As per the original commit, different countries implemented the lookup differently. Therefore the results may have to be amended on a country per country basis. The original commit works perfectly for the UK (and for Europe probably). Not inclined to accept a patch that breaks functionality I need.. But I appreciate that USA based users require different functionality. Therefore id reject this, but with a request to implement a enhancement to have some per-country customisations of the results returned, with the USA and UK pre-populated.

smackjack3:
Sorry about that - I didn't see the original commit or realize it varied.  Just made some changes to store the location fields in variables during the foreach so we can pass either the level_1 or level_2 value depending on whether it's a US-based location after the loop ends.  Thought I was editing my fork but apparently I was editing the commit so hopefully it looks alright!

P.S. I'm a fairly novice developer with little git experience (mostly my own private repos on Gitlab), so bear with me!  I'm starting to use openCATS at my company so looking forward to help with any development work that I can though.

RussH:
Hi @smackjack3 no problem, my fault, I should have called out that more clearly in the release notes. Have asked some of the devs to review & comment.  Please continue your good work and I hope you're enjoying opencats!

RussH:
update to the location mapping - thanks @smackjack3 !

PR #250, opened 2017-04-26
Remove references to CATS ASP module which is not present in OpenCATS

### LINKED ISSUE #247: Mass Import not working in 0.9.4 (testing)
Mass import displays the wrong page in 0.9.4 (testing).

mlespiau:
Remove references to modules/asp modules and dummy file. 
Fixes #247


mlespiau:
@skrchnavy please review :)

modules/settings/SettingsUI.php
```diff
@@ -681,10 +672,6 @@ public function handleRequest()
                 $this->getFirefoxModal();
                 break;
 
-            case 'downloads':
```
skrchnavy: Is this really not used in non-asp part? I didn't check all code so just assure we are not removing more then necessary.

mlespiau: I did not find any link to this page in the whole code base.

modules/settings/SettingsUI.php
```diff
@@ -2456,7 +2443,7 @@ private function administration()
         $careerPortalUnlock = false;
         $careerPortalSettings = new CareerPortalSettings($this->_siteID);
         $cpData = $careerPortalSettings->getAll();
-        if (intval($cpData['enabled']) || (file_exists('modules/asp') && !$_SESSION['CATS']->isFree()) ||
+        if (intval($cpData['enabled']) || !$_SESSION['CATS']->isFree() ||
             LicenseUtility::isProfessional())
```
mlespiau: file_exists('modules/asp') always evaluated to true - I just provided equivalent code at this point while removing the reference to modules/asp file.

skrchnavy: I mean `LicenseUtility::isProfessional()` is valid check? Do we need it in the code of OPEN version?

mlespiau:
We don't need it in the code but we should clean it up in another commit. I tried to that in the past but was not that simple.

PR #252, opened 2017-04-27
Email templates user signature

skrchnavy:
Added an e-mail signature variable into e-mail templates, which is configurable for each user and from the GUI under My Profile settings (so it is available to everyone, not just admins).
It utilizes the CKEditor.

Limitation: 255 chars might be enough for plain text, but might not be for HTML formatted text - this could be extended but schema update is required.

Example of email signature:   
> My Name | 0900 200 300 | <a href="http://www.myWebsite.com">myWebsite.com</a>  
> <a href="mailto:myEmail@gmail.com">myEmail@gmail.com</a>


RussH:
this is now targeting a php7 feature branch. It was originally on the php5 master branch. Do I need tor rebase & merge if/when accepting it?

shantadam:
My observations:
1- Indeed, it's been couple of years held behind. It certainly needs a rebase.
2- I concur with Codacy's red flagged issues. However, the project is infected with bad coding habits and free will usage of the global variables. 
3- I also concur with Codacy's unnecessary usage of "else" statement.

My recommendation: 
1- Since, there has been none enforcing the best coding practice, I'm tempted to push for the implementation of a singleton util class object, once and for good, to resolve the references to the required global variables. This project needs to get in shape.

shantadam (approved):
I won't hold back the progress by approving this feature. However, we'll have to consider implementing getters for global variables and being more proactive on applying coding best practices.

PR #229, opened 2017-03-30
Ckeditor font plugin only where needed

skrchnavy:
font plugin in ckeditor is not for in common ckeditor config but is added into CKEDITOR instance creation.

PR #257, opened 2017-05-01
Ckeditor in composer

skrchnavy:
composer used to 'download' ckeditor. imports updated to uses vendor/ckeditor.

version used: 4.5.1 as was upgraded in #132


skrchnavy:
PR #229 removes changes in config.js of ckeditor, so that PR shall be approved first and then this PR.
Or when this PR applied, conflicts in the other PR have to be resolved (config.js file)

PR #258, opened 2017-05-01
Make backup resilient to missing files

### LINKED ISSUE #253: I think the GUI backup is broken
I'm trying to do the documentation for backup/restore and I can't get it to work in 0.9.3.  Not sure if it's a code change or some sort of permissions change.

The GUI backup is in the admin page, site backup.
It backs up the database and the resumes.  When you click it, it performs a full, restorable backupfile (backup.bak) which can be downloaded through a link after.  It also offers the option of breating an attachment-only backup.
This morning, I tested and the full backup in 0.9.3 doesn't work, but the attachment backup does.  I also tested in 0.9.1a and it does work there.

My initial thinking is that maybe some sort of permissions issue that OpenCATS can't do this, but again, it could be the code changes too.

mlespiau:
Backup is failing on some instances when an attachment is on the db but it's missing in the file system.

This PR makes the backup code more resilient to this issue and proceed with the backup ignoring this file.

We should also make a tool to remove orphaned attachments #259. 

Fixes #253

### DIFF
--- a/modules/settings/ajax/backup.php
+++ b/modules/settings/ajax/backup.php
@@ -239,6 +239,23 @@ if ($action == 'backup')
     while ($row = mysql_fetch_assoc($queryResult))
     {
         ++$attachmentCount;
+        $relativePath = sprintf(
+            'attachments/%s/%s',
+            $row['directory_name'],
+            $row['stored_filename']
+        );
+        if (!file_exists($relativePath)) {
+            setStatusBackup(
+                sprintf(
+                    'Skipping attachment that\'s missing on the file system: %s (%s of %s files processed)...',
+                    $relativePath,
+                    $attachmentCount,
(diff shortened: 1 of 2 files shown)


modules/settings/ajax/backup.php
```diff
@@ -239,6 +239,23 @@ function setStatusBackup($status, $progress)
...
+                    $relativePath,
+                    $attachmentCount,
+                    $totalAttachments
+                ),
+                ($attachmentCount / $totalAttachments)
+            );
+            continue;
+        }
 
         setStatusBackup(
```
skrchnavy: @mlespiau I propose to add $relativePath info also into this `setStatusBackup` we have value prepared and it could align rows.

No further comments.

skrchnavy:
No more comments from me

PR #260, opened 2017-05-01
Legacy root added to includes

skrchnavy:
this PR is to prepare code for migration to symfony.


skrchnavy:
Some files listed above updated (few more where I found another include). It is not 100% covered, don't think it is necessary, we need to reduce number of changes in symfony branch and this PR shall help.

PR #263, opened 2017-05-04
i18n

virtimus:
i18n based on php-gettext. 
Polish version added. 
Should be 90% complete.

set some env variables:

```
set ATS_WORKSPACE_DIR=[root OpenCATS sources]
set GETTEXT_HOME=[path to gettext binaries]
```

and use ant tasks in build.xml to:

gt-generateFileList - generate list of files to be included in gettext
gt-run-xgettext - to regenerate pot file based on output from gt-generateFileList
gt-run-msgmerge-pl - to merge pot into Polish po file


skrchnavy:
Please revert some of config.php changes to previous state,some defaults were changed and these changes are not related to i18n.

PR #273, opened 2017-06-03
Php 5.5 support

skrchnavy:
updated const arrays to static class methods to be compliant with php 5.5


skrchnavy:
this is slightly in conflict with #290 due to the fact, that ACL config was in #290 extracted to new file. so based on order of merging, manual conflict resolution have to be done.

PR #291, opened 2017-06-28
travis build fix in v0.9.4

skrchnavy:
Due to changes on travis and docker images.
docker-engine is not available, docker-ce shall be used instead


mlespiau:
thanks @skrchnavy for the fix!

PR #340, opened 2017-11-27
Fix two minor bugs

wjcheers:
1. there is a bug in the time format while editing activity.
strtotime("9:0 AM"); output nothing. It lead the new date of activity to
00:00:00.

2. sortBy 'Description' is a typo. the correct one is sortBy 'description'


RussH:
I see no issues with accepting this! .. thank you @wjcheers

mlespiau:
Thanks @wjcheers

PR #357, opened 2018-03-26
Fix insecure SMTP

p-rintz:
A past phpmailer update changed to assume to use TLS even if its not enabled. This breaks insecure SMTP connections.


RussH:
Should we really be fixing this by pulling in the latest version of phpmailer instead?

p-rintz:
A phpmailer update would not fix this,as its still the same in even the newest versions.

https://github.com/PHPMailer/PHPMailer/blob/da9fede584e200038ae78e7cd153d092a210f9b7/src/PHPMailer.php#L263

RussH:
Thank you!

PR #366, opened 2018-06-30
fix LDAP module

wumb0:
I got this working properly with both Samba (AD) and OpenLDAP. LDAP and LDAPS.  
Issues preventing it from functioning:  
1. `LDAP_OPT_REFERRALS` may be required for samba or AD. Added an option to config.php to set this if desired.  
2. The second `ldap_bind` in the __authenticate__ function (after the user is found by the LDAP_BIND_DN user) uses `$result[0][LDAP_ATTRIBUTE_DN][0]` as the binding DN. This gets the first character of the `LDAP_ATTRIBUTE_DN` attribute of the first result of the query and not the desired attribute. I removed the [0] to resolve the DN correctly.  

Cosmetic fixes:  
1. Removed mix of tabs and spaces.   
2. Added some error logging.  
3. Added a clearer example of how the `LDAP_ACCOUNT` config setting is supposed to be used.

### DIFF
--- a/config.php
+++ b/config.php
@@ -272,7 +272,7 @@ define ('LDAP_BASEDN', 'dc=example,dc=com');
 define ('LDAP_BIND_DN', 'cn=read-only-admin,dc=example,dc=com');
 define ('LDAP_BIND_PASSWORD', 'password');
 
-define ('LDAP_ACCOUNT', '{$username}'); // '{$username}' cannot be changed, else can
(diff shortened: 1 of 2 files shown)


blazej-czuk:
@wumb0 Are you sure this patch does not introduce any issues?
Could you please help me identify the problem? 
https://github.com/opencats/OpenCATS/issues/530

Issue #9, opened 2015-01-13
Candidate tab and data error


adnanalitribe:
i using cats-0.9.2 for localhost (and using WAMP 2.0) , its install successfully and when click on candidate tab its show an error "Query Error -- Report to System Administrator ASAP" with some detail.  below i post screen shot of it
![2015-01-14_003044](https://cloud.githubusercontent.com/assets/10214824/5727656/93a44bd8-9b86-11e4-9b34-bff266cd70cf.png)

please hlp me to resolve it. i also have cats-0.9.1 and install it .. old version working fine no error in it

RussH:
Hi,

I remember this was a problem when the tagging feature was introduced...
but thought it was fixed. Did you install 0.9.2 directly from the github
snapshot?

Russ

RussH:
Yes, definitely a problem, i can fix by copying ove rthe code from 0.9.1

adnanalitribe:
Over the code , a code from 0.9.1 ... (RussH , please can you guide me from where i found that code , from which files [or line number] ) thanks for response on my error

RussH:
Hiya, Okay, I can see where the problem is but will leave it for another night.. If you select 'Install a blank database' the correct install routine is used (but you need to use admin/admin to log in afterwards) - if you select 'Install test data' then the sql db isn't updated correctly, and the tagging table isn't generated. 

So - the immediate workaround is to choose the 'create a blank database' option, I will fix the 'test data' sql routine.

RussH:
finally - the 'test data' is pulled from ./db/cats_testdata.bak whereas the production blank db is from cats_schema.sql. So I need to update cats_testdata.bak

adnanalitribe:
Thanks Russh , i will reinstall it with empty database , (new) ,

alexandre433:
Sounds like fixed on #400 should be closed

PR #398, opened 2018-10-30
Update db/cats_testdata.bak

RussH:
Updated test data for MySQL;
- strict mode compliant (now the default), and;
- added in candidate_tag tables to MySQL table structure

Installer does not error now when you select the demonstration data.


RussH:
resolves issue #9

PR #401, opened 2018-10-30
optional update; iCal feed

RussH:


RussH:
you're quite right, this will never go in, it was a placeholder to retain the non-github contribution. It's held as a Github issue for now - btw I do run this myself, it works great.

shantadam:
@RussH, :) 
In fact, the goal is to simplify and automate. Which is why Outlook synchronization should be integral part of OpenCATS feature. I'm working on it. 
iCalendar/iCal (Internet Calendar) can be discouraging to the business users. Since it involves manual configuration. 
P.S.: Expect efficiency feature contribution from my end soon, which optimize productivity. Work is in-progress.

PR #440, opened 2019-06-28
Address vulnerabilities

RussH:
To address the XXE vulnerability raised by Mark Ruther.

### DIFF
--- a/.github/no-response.yml
+++ b/.github/no-response.yml
@@ -1,7 +1,7 @@
 # Configuration for probot-no-response - https://github.com/probot/no-response
 
 # Number of days of inactivity before an Issue is closed for lack of response
-daysUntilClose: 14
+daysUntilClose: 16
 # Label requiring a response
 responseRequiredLabel: more-information-required
 # Comment to post when closing an Issue for lack of response. Set to `false` to disable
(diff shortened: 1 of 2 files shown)


shantadam:
Can we Kindly request to get this important fix part of test-php-7.x too?

RussH:
well the intention would normally be that test-php-v-7.x would be merged
back into master prior to release - but if you're actually building against
test, then sure I'll add it in there, too


----
Russ

shantadam:
indeed, Russ, I am building on top of test-php-v-7.x.

If you can give me a hand with the security fix merge to the later branch, you would help me focus on advancing enhancements in that branch.

Much appreciated, Russ. Thanks.

Regards,
Shant Adam
--
DISCLAIMER - This message contains confidential information and is intended only for the individual named. If you are not the named addressee you should not disseminate, distribute or copy this e-mail. Please notify the sender immediately by e-mail if you have received this e-mail by mistake and delete this e-mail from your system. E-mail transmission cannot be guaranteed to be secure or error-free as information could be intercepted, corrupted, lost, destroyed, arrive late or incomplete, or contain viruses. The sender therefore does not accept liability for any errors or omissions in the contents of this message, which arise as a result of e-mail transmission.

Issue #444, opened 2019-09-11
rss is not valid


ThomasKujawa:
### Expected behavior and actual behavior.

Expected: valid
actual: not valid

### Steps to reproduce the problem.

https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Fwww.personal247.de%2Frss%2F

### What version of opencats are you running? 

0.9.4

### What version of PHP and MySQL are you running

Server:  via TCP/IP
Server-Typ: MySQL
Server-Version: 5.7.26-nmm1-log - (Ubuntu)
Protokoll-Version: 10
Benutzer: @85.13.144.137
Server-Zeichensatz: cp1252 West European (latin1)

Apache
Datenbank-Client Version: libmysql - mysqlnd 5.0.12-dev - 20150407 - 
PHP-Erweiterung: mysqliDokumentation curlDokumentation mbstringDokumentation
PHP-Version: 7.2.19-nmm1

phpMyAdmin Versionsinformationen: 4.9.0.1

RussH:
PHP 7.x is not supported on the 0.9.4.x release version and will generate errors

shantadam:
Hello Thomas,

I’m currently helping @RussH on the app enhancement and bug resolutions.

PROBLEM:

  *   Further peaking in your error log, I noticed the message clearly state that a non-static function is being called as if it was declared to be static.
  *   This is certainly a bug. Which need to be resolved.
  *   The following files reference to a non-static function:
     *   OpenCATS/modules/rss/RssUI.php (line 93 and line 124)
     *   OpenCATS/modules/xml/XmlUI.php (line 97 and line 163)

ANALYSIS:

  *   An object class typically declare non-static function(s), which get reference calls through instantiation, as such::
     *   class DateUtility() { public function getRSSDate(){…} }
     *   $dateUtil = new DateUtility();
     *   $dateUtil->getRSSDate();
  *   Optionally, an object class may declare static function(s), which get reference calls directly. The later functions are independent of class variable initialization, as such:
     *   class DateUtility() { public static function getRSSDate(){…} }
     *   DateUtility::getRSSDate();
  *   In this case, some programmer referenced the non-static function as if it was static call.

RESOLUTION:

  *   Further analyzing the project source code for the use of getRSSDate(), I can confirm that the call should be made statically, as CATSUtility::getAbsoluteURI() is done.
  *   To resolve the issue we need to correct the function’s signature declaration in OpenCATS/lib/DateUtility.php, as such:
[...]

PR #447, opened 2019-09-27
Resolved issue #444 - rss is not valid

shantadam:
- Further analyzing the project source code for the use of getRSSDate(), it is confirmed that the call should be made statically, as CATSUtility::getAbsoluteURI() is done.
- To resolve the issue we needed to correct the function’s signature declaration in OpenCATS/lib/DateUtility.php, as such:
- Added the “static” property at the beginning of the function signature line 350, as such:
- static public function getRSSDate($unixTime = false)


RussH:
Thank you Shant.. appreciated!

PR #491, opened 2020-09-10
update travis.yml after yml validation

RussH:
remove spurious hyphen in travis.yml


RussH:
Looks like there's a further syntax issue in mailer.php. Will propose a fix for that & resubmit

PR #498, opened 2020-10-02
Update careers/index.php

pribeiro42:
The LEGACY_ROOT constant is defined in config.php but this file is not included in the careers/index.php so, there's an error, resulting on a blank page showed to the user.
The inclusion of config.php in this file solves this problem


shantadam (approved):
Good catch.

RussH:
thank you!

PR #526, opened 2021-02-19
Include field id for the address field.

### LINKED ISSUE #356: input-address req - Field Definition Missing ID
### Expected behavior and actual behavior.

When I mark `<input-address>` as required: `<input-address req>` the field is not checked and the submission of an application will not be processed.  Upon further review, code for the placehoder replacement with input field is entering the address field without its ID.  Therefore, this breaks the Javascript which is looking for id="address" 

- need to add id="address" to both areas of the code to fix this bug.

pribeiro42:
Fix #356 by including the field id when replacing the placeholder code


RussH:
Thank you @pribeiro42 !

PR #532, opened 2021-04-06
Update LDAP.php

blazej-czuk:
This change makes the attribute mapping correct for me.


RussH:
Thank you @blazej-czuk !

PR #552, opened 2021-09-24
Fixed critical upload vulnerability

Nickguitar:
There is a **critical vulnerability** that allows **remote code execution** due to lack of proper validation of the uploaded file.

This affects all versions of OpenCATS I could test. (Editing .htaccess does not fix the issue in non-Apache webservers)

Just to show how critical this is, I wrote a simple script to exploit this vulnerability.

I did not publish this script yet, but I will do so within 90 days from today or right after this issue has been fixed (either via this pull request or otherwise).

The result against both Linux and Windows webservers was this:

image

image

### DIFF
--- a/lib/FileUtility.php
+++ b/lib/FileUtility.php
@@ -184,11 +184,17 @@ class FileUtility
 
         /* Is the file extension safe? */
         $fileExtension = self::getFileExtension($filename);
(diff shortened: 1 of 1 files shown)


RussH:
Thanks for the investigation and proposed fix Nick, will review and produce a secure release

RussH:
Thanks to @Nickguitar for the investigation and fix

Issue #466, opened 2020-02-28
Candidate Not Registered


grichardomi:
Using Firefox,

Login an existing candidate from the Career Page, I'm getting this message "You have not registered yet.  Please wait while we direct you to the job list"

Can anyone help with the cause of this message or how to register a candidate?

Thank you!

RussH:
so I don't use the careers page, however;

https://demo.opencats.org/careers/

should reproduce what you want. If you haven't registered, you just browse the job listings, and then when you try to apply for one, you're prompted to  register.

grichardomi:
Thanks for responding.

I still get the message after registering (admin@example.com | test4 | 84770) . And it is not allowing me to update my profile.

shantadam:
Hi grichardomi,

Forgive me for going with your username, but I have no reference to your first name.

As for the issue, may I ask what is your role while accessing OpenCATS? Were you given access to a link? Are you a user? a candidate? An admin? If not an admin, do you have an admin? Are you trying to install? Is it an existing system?

We need some more in-depth knowledge of the context on who is accessing and how. Simply mentioning browser as Firefox, won’t help.

Regards,
Shant Adam,


On Feb 28, 2020, at 12:09 PM, grichardomi <notifications@github.com> wrote:

﻿

Using Firefox,

Login an existing candidate from the Career Page, I'm getting this message "You have not registered yet. Please wait while we direct you to the job list"

Can anyone help with the cause of this message or how to register a candidate?

Thank you!

grichardomi:
what is your role while accessing OpenCATS? 
Candidate

Were you given access to a link?
https://demo.opencats.org/careers/
admin2@example.com | lastname: test5 | zipcode: 84770 

 Are you a user?
Candidate

 If not an admin, do you have an admin? 
Candidates should not have to have admin access in order to update their profile

Are you trying to install? 
I experience same behavior on demo site and on an newly installed server.

shantadam:
I only have good news for you from my end. All my tests, through your provided link along with the candidate credentials passed successfully by logging in.

At this point, I have two favours to ask you:
- Please find the way to clear your specific browser’s cache.
- If issue persist, please provide your browser’s and OS detail specs. Not just names.

Regards,
Shant Adam


On Feb 29, 2020, at 5:03 PM, grichardomi <notifications@github.com> wrote:

﻿

what is your role while accessing OpenCATS?
Candidate

Were you given access to a link?
https://demo.opencats.org/careers/
[...]

PR #578, opened 2022-09-25
Fix "Candidate not registered" error and security fixes

hansmach1ne:
This commit fixes "Candidate Not Registered" issue #466.


RussH (approved):
The regex for candidate not registered checks out, and the changes to restrict XSS on internal pages are very much apprecited, this closes internal XSS vulnerabilities as well as external.

PR #581, opened 2022-10-03
SQL injection fix in Tag update functionality

hansmach1ne:


RussH:
Thank you for the fix @hansmach1ne

Issue #618, opened 2023-05-25
Job Orders always converted to HTML code for viewing (instead of readable text)


zoomiest2:
### Expected behavior and actual behavior.

- Inserting text into a new Job order, results in text being converted to a blob of HTML code for viewing.

### Steps to reproduce the problem.

- Create a new Job Order. Paste in formatted or unformatted text. Its always converted to a blob of html.

### What version of opencats are you running? 
- 9.7.2

#### Release or downloaded from Git? 
- OpenCATS-v0.9.7.2-FULL.zip (full stable version)

#### WAMP or LAMP? 
- LAMP

### What version of PHP and MySQL are you running
- PHP Version 7.2.34-38+ubuntu22.04.1+deb.sury.org+1
- MariaDB 10.6.12

RussH:
HI Zoomiest!

Sorry for no response - ﻿am busy mostly working with the PHP 8.2 version I'm slowly improving / trying to debug. I'll get the 9.7.2 version and php 7.2 installed and see what's happening.

sm-programmer:
Hi there!

It seems that the function `getSanitisedInput` is being applied to most fields, in order to limit XSS scripting attacks.

In particular, it is being applied to the `description` field, which applies `htmlspecialchars` over it and converts any tags to its corresponding entities, which in turn mangles the HTML and renders it as plain text.

Reverting to using `getTrimmedInput` works, of course, but that also opens the possibility of XSS attacks over the `description` field.

What can it be done to overcome this issue?

* I have read that whitelisting can be used, but it is quite the effort, as there are many attacks possible that do not use `<script>` tags explicitly.
* Is using another markup language being considered? For example, Markdown.
* I have also come across the idea of saving both the unprocessed and processed variants of the string in DB. The problem is that the editor itself interprets the HTML being stored, which is yet another XSS attack vector.

Any thoughts in this matter are greatly appreciated.
Regards!

RussH:
Thanks for the investigation﻿.. Yes it will of course getSanitisedInput. There are options for testing the options on htmlspecialchars - such as https://onlinephp.io/htmlspecialchars - perhaps you can test ot see if we can reconfigure this field to be less strict, but still provide XSS protection?

RussH:
seems to be for CKEditor (rather than a plain 'ol html form), CKEditorneeds to be configured to use something like DOMPurify;
https://ckeditor.com/docs/ckeditor5/latest/features/html/html-embed.html
[...]

PR #621, opened 2023-06-13
Revert htmlspecialchars for CKEditor

RussH:
Revert the security restrictions deployed with HTMLSPECIALCHARS for description fields using CKEditor, as it strips all HTML formatting


RussH:
fixes #618

PR #534, opened 2021-05-11
Security fixes

RussH:
- Partial mitigation for CVE-2021-25294

(change serialization / unserialization to use JSON - POC patch as submitted)

- Mitigation for CVE-2021-25295

(apply restrictive htaccess to /attachments and /upload) - and guidance in the project wiki.

Thanks go to Security Researchers Jens Winkle and Simon Jungherz for identifying these issues and reporting the vulnerability.

PR #631, opened 2023-10-01
Sync allowed format lists in FileUtility.php and .htaccess files

xalt7x:
Apply FileUtility.php changes from PR #576 to .htaccess files from PR #534


RussH (approved):
Looks good, adds a few file formats to the regex.

PR #641, opened 2024-04-16
RussH patch cookies

RussH:
update session_cookie to include samsite, httponly options. Resubmitting after some Travis-CI issues.


RussH:
this patch uses syntax only for PHP7.4. need to revisit

PR #649, opened 2024-05-20
Add "txt" to the list of safe file extensions

xalt7x:
so that text files will not have a duplicate extra extension


xalt7x:
Somehow [I missed](https://github.com/opencats/OpenCATS/commit/b1af3bde1f68bec1c703ad66a3e390f15ed8ebe1#commitcomment-80485043) adding "txt" to the list of "GoodFileExtensions"...

RussH:
Okay, all added!

PR #650, opened 2024-05-20
Fix javascript "back button" links

xalt7x:
Use history.back() method instead of currently non-functional back()


RussH:
Thanks - good catch!

PR #651, opened 2024-05-21
Use mb_substr() for correct abbreviation of non-ASCII characters

xalt7x:
When using substr() or another method to reduce a string to/by 1 byte, many UTF-8 characters are lost (displayed as � ). Switching to mb_substr() fixes this.

### DIFF
--- a/lib/Candidates.php
+++ b/lib/Candidates.php
@@ -2066,7 +2066,7 @@ class CandidatesDataGrid extends DataGrid
                                      'filter'         => 'candidate.web_site'),
 
             'Key Skills' =>    array('select'  => 'candidate.key_skills AS keySkills',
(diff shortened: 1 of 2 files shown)


xalt7x:
The problem is easily reproducible with Cyrillic/Ukrainian characters (e.g., "Джон Дое" as the User/Owner name, or "Навички обслуговування клієнтів" string for "Key Skills").

fix_cyrillic_abbreviation


[Additional information](https://stackoverflow.com/questions/14785682/substr-doesnt-work-fine-with-utf8/14785709#14785709):

> If you’re working with strings encoded as UTF-8 you may lose characters when you try to get a part of them using the PHP substr function. This happens because in UTF-8 characters are not restricted to one byte, they have variable length to match Unicode characters, between 1 and 4 bytes.

RussH:
Thanks @xalt7x !

PR #652, opened 2024-05-28
joborders: for forms, take a default company name from the database

xalt7x:
Default company name is saved in the database and can be changed. However, for "Add Job Order" and "Edit Job Order" forms it was hardcoded to the standard name ("Internal Posting").

default-company-name


RussH:
Thank you for the commit!

PR #660, opened 2024-09-16
sync changes in  master to branch

RussH:


RussH:
will add php 8.x checks to master and re-sync

Issue #663, opened 2025-07-07
Session error : Empty page after login page


meorajrul:
Machine information:

1. OS : MacOS 15.5
2. Browser : Safari 18.5
3. Host method : Laravel valet/herd
4. PHP Version: 7.4
5. Database Version : 8.0.33
6. Web Server : Nginx 1.29

OpenCATS version : 0.9.7.4 (0c1487b931755c7a282057ff690eada4a46ff664)

Step to produce:

1. Git clone master branch
2. Do setup wizard till finish
3. Login
4. Error output with empty content
`
Warning: Cookie paths cannot contain any of the following ',; \t\r\n\013\014' in /Users/xxxx/project/www/OpenCATS/lib/Session.php on line 900
`

<img width="1504" alt="Image" src="https://github.com/user-attachments/assets/5e72e42f-367e-4c80-bfc4-69d2828fef8e" />

RussH:
Note that php 7.4 isn't a supported codelevel for this release. Max of PHP 7.2, please re-assess using the PHP 8.2 branch. have accepted the PR though!

PR #664, opened 2025-07-07
fix session error when domain are not set, default to empty

meorajrul:
This is to address session error related to issue #663


RussH:
@meorajrul  thank you for the commit. Note most development is happening on the php8.2 branch.

PR #675, opened 2025-12-05
feat: introduce address2 and replace address textareas with inputs

anonymoususer72041:
## Summary

Replace textarea-based address fields with single-line `<input>` fields across the main UI and the careers portal and add “Address Line 2” (`address2`) support for candidates, contacts, and companies.

Add schema upgrades for existing installations and migrate legacy multi-line address values into `address` (line 1) and `address2` (remaining lines).

## Motivation

A large multi-line textarea often implies "enter a full postal block here" while many users (depending on country) expect a single street line. At the same time, international addresses commonly require an additional line (suite/floor/building).


anonymoususer72041:
It looks like the fix for #356 (merged into develop via #526) never made it into master, so this Pull Request includes the missing `id="address"` behavior again.

Issue #344, opened 2017-12-15
Day/Month mixed up when editing activities if UK locale set


Mephisztoe:
### Expected behavior and actual behavior.
When configured to use UK locale instead of US for datetime values (DD-MM-YYYY instead of MM-DD-YYYY), I expect an activities' month and day to be displayed correctly when editing an entry on a candidates' page. Instead they are mixed up.

### Steps to reproduce the problem.
- Set locale to UK
- Log an activity on a candidates' page (date is e.g. August, 4th)
- Click on edit button for activity

Result: 
The month Combo Shows up April, the day Combo 8th.

### What version of opencats are you running? WAMP or LAMP? 
XAMPP on Windows.

- Operating System: | Windows NT ******** 6.2 build 9200 (Windows 8 Enterprise Edition) i586
- Operating System Type: | CATS thinks your operating system is Windows.
- PHP Version: | 5.6.28
- Database Version: | MySQL 10.1.19-MariaDB

skrchnavy:
Tested on demo page where is date format set to UK:
![image](https://user-images.githubusercontent.com/7057580/34071885-018f9cec-e27e-11e7-9888-6c98179e17fc.png)

skrchnavy:
Error probably caused in file: https://github.com/opencats/OpenCATS/blob/00166b4c3c83d0dcc9b7627411b144cb0d77758c/js/activity.js#L277
```
        dateSpan.innerHTML = DateInputForDOM('dateEditActivity' + activityID, true, 'MM-DD-YY', dateAndTime.substr(0,dateAndTime.indexOf(' ')), -1);
```
where is hardcoded format to `'MM-DD-YY'`

PR #683, opened 2025-12-08
fix: respect user date format for activities, calendar events and core date fields

anonymoususer72041:
## Summary

This PR makes OpenCATS consistently respect the **user date format** (`MM-DD-YY` vs `DD-MM-YY`) across the main UI entry points for dates, while keeping the database storage unchanged (`YYYY-MM-DD` / `YYYY-MM-DD HH:MM:SS`).

Concretely:

- **Activities (candidates & contacts)**
  - Use a shared `window.CATSUserDateFormat` (`MM-DD-YY` or `DD-MM-YY`) for:
    - Editing existing activities (fix for #344).
    - Creating new activities via the activity/change status and schedule event modals.
  - On the server side, validate and convert activity dates using a `dateFormatFlag` derived from `$_SESSION['CATS']->isDateDMY()`, before converting to `YYYY-MM-DD HH:MM:SS`.

- **Calendar events**
  - Use the user date format when:
    - Rendering `DateInput` fields for `dateAdd` and `dateEdit`.
    - Populating those fields via `SetDateInputDate(...)` in `CalendarUI.js`.
  - Parse and validate `dateAdd`/`dateEdit` using the same user date format flag in `CalendarUI.php`, then convert to `YYYY-MM-DD` for storage.

- **Candidate availability (`dateAvailable`)**
  - Render `dateAvailable` using the user date format on add/edit.
  - On submit, validate against the user date format and convert to `YYYY-MM-DD` before saving.

- **Job order start date (`startDate`)**
  - Render `startDate` using the user date format on add/edit.
  - On submit, validate against the user date format and convert to `YYYY-MM-DD` before saving.

- **Extra fields of type “Date”**
  - Introduce a helper that determines whether the user/site uses `DD-MM-YY` or `MM-DD-YY`, working both for logged-in users and the careers portal.
  - Render date extra fields using `DateInput(..., userDateFormat, ...)`.
  - When editing:
    - Convert the stored canonical value to the user date format for display.
    - Validate the posted value against the user date format.
    - Convert back to a canonical `MM-DD-YY` representation for storage in the extra field value.

Across all these changes:
- The **user-facing date format** is now consistent with the user/site setting.
- The **database storage format remains unchanged** (`YYYY-MM-DD` / `YYYY-MM-DD HH:MM:SS`), and only the parsing/validation layer was updated.

## Motivation

Previously, OpenCATS assumed `MM-DD-YY` in several places in the UI and server-side validation, regardless of the user’s configured date format. This led to a few concrete problems:

- **Bug in activity editing (#344)**:  
  When the user date format was set to `DD-MM-YYYY`, existing activity dates were still parsed as `MM-DD-YY`. As a result, day and month were swapped when editing (e.g. `04-08-25` intended as 4 August became 8 April when saved).

- **Inconsistent UX when entering dates**:  
  Some forms always rendered `DateInput(..., 'MM-DD-YY', ...)` even when the rest of the UI (and the user’s expectation) used `DD-MM-YY`. This made it easy to enter a “valid-looking” date that was later misinterpreted on the server.

- **Risk of silently wrong dates in the database**:  
  Because the database expects `YYYY-MM-DD`, but the conversion step assumed `MM-DD-YY`, users with `DD-MM-YY` could end up storing logically incorrect dates (day/month swapped) without any error.

This PR addresses these issues by:

- Using a single source of truth for the **user date format** (`isDateDMY()` / site preference for unauthenticated users).
- Passing that format down into all relevant `DateInput`/`DateInputForDOM`/`SetDateInputDate` calls.
- Validating and converting user input according to that format before turning it into `YYYY-MM-DD` / `YYYY-MM-DD HH:MM:SS` for storage.

The goal is to make date handling predictable and safe for both `MM-DD-YY` and `DD-MM-YY` users, without changing the existing database schema or storage format.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,128 @@
+name: CI/CD Pipeline
+
+on:
+  push:
+    branches: [ master, develop, feature/** ]
+    tags: ['v*']
+  pull_request:
+    branches: [ master, develop ]
+  workflow_dispatch:
+
+jobs:
+  tests:
+    name: Tests (PHP ${{ matrix.php-version }})
+    runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      pull-requests: write
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version: ['7.2']
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          tools: composer:v2
+
+      - name: Cache Composer packages
+        uses: actions/cache@v4
+        with:
+          path: ~/.composer/cache
+          key: ${{ runner.os }}-php-${{ hashFiles('**/composer.lock') }}
+          restore-keys: ${{ runner.os }}-php-
+
+      - name: Install Dependencies
+        run: composer install --no-progress --prefer-dist
+
+      - name: PHP Syntax Check (Lint)
+        run: find src -name "*.php" -print0 | xargs -0 -n1 php -l
+
+      - name: Composer Security Audit
+        # We use || true so legacy vulnerabilities don't block the build
+        run: composer audit || true
+
+      - name: Create Reports Directory
+        run: mkdir -p reports/behat-default reports/behat-security test/screenshots
+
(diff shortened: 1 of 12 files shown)


anonymoususer72041:
Currently this PR introduces very strange behavior in my productive system. Editing an activity unexpectedly changes the date (e.g., the year updates to today's date).

PR #685, opened 2025-12-14
security: switch password storage to password_hash() with legacy MD5 migration

anonymoususer72041:
## Summary

This pull request replaces the legacy MD5-based password handling with PHP's `password_hash()` and `password_verify()` functions using PASSWORD_DEFAULT and ensures that newly set or changed passwords are always stored using the modern hashing API.

For existing installations, legacy MD5 hashes are still accepted at login but, on successful authentication, the plaintext password is rehashed using PASSWORD_DEFAULT and the stored hash is updated. For modern hashes, `password_needs_rehash()` is used after a successful login to transparently upgrade hashes if PASSWORD_DEFAULT changes in the future.

For new installations, the base schema now defines `user.password` as `VARCHAR(255)`, the installer’s schema upgrade path adds an `ALTER TABLE` step so existing databases are migrated to the same definition and the Security.MD documentation has been updated to reflect that OpenCATS now uses `password_hash()` and `password_verify()` with PASSWORD_DEFAULT for password storage.

## Motivation

MD5 is no longer considered appropriate for storing user passwords, and modern PHP provides a well-established and actively maintained API for password hashing via `password_hash()` and `password_verify()`, with PASSWORD_DEFAULT acting as a safe, forward-compatible algorithm selector that can adopt stronger algorithms over time without requiring further application changes.

Integrating `password_needs_rehash()` independently of the MD5 compatibility logic ensures that future changes to PASSWORD_DEFAULT or its options can be adopted at runtime without another invasive refactor and increasing the password column to `VARCHAR(255)` while updating Security.MD keeps schema and documentation aligned with current best practices and avoids further schema changes when PASSWORD_DEFAULT eventually moves to a stronger or longer hash format.


anonymoususer72041:
CI/CD Pipeline is currently failing because this PR also reactivates the First-Login-Wizard if default credentials are set.

However, /test/features/login.feature currently assumes that the dashboard will load after login.

@RussH seems like there is a bigger refactor needed. Alternatively we can initially just deactivate the restored First-Login-Wizard revisit this later if you are okay with this.

anonymoususer72041:
> We need the follow-up wizard fix committed ASAP; if it doesn’t land before the next release, we’ll have to revert this PR to avoid shipping with a known default-password path.

The First-Login-Wizard was already broken before – it just got visible while working on this PR. Even if this commit will be reverted, the First-Login-Wizard will still not be triggered.

PR #689, opened 2025-12-27
Normalize initial database state for new installations

anonymoususer72041:
## Summary

This PR normalizes the initial database state created from `db/cats_schema.sql` by removing demo/sample tag data, resetting instance-specific seeded values to deterministic defaults, and replacing `testdomain.com` placeholders with the IANA-reserved placeholder domain `example.com`.

In addition, it updates `modules/install/Schema.php` to migrate existing databases where `site.last_viewed_day` is `NULL` or `0000-00-00` by backfilling a sentinel date (`1000-01-01`) and then enforcing a non-null default at the schema level.

## Motivation

Fresh installs should start from a clean, deterministic baseline without embedded demo content or leftover environment-specific values. This reduces confusion, makes the initial schema easier to reason about, and avoids noisy diffs when comparing or exporting databases.

Backfilling and normalizing `site.last_viewed_day` prevents `NULL`/zero-date edge cases and aligns this field with the sentinel-date approach used elsewhere in the schema while keeping existing instances consistent through an explicit migration step.

Replacing `testdomain.com` with the IANA-reserved placeholder domain `example.com` avoids potential confusion with real domains.


anonymoususer72041:
@RussH any chances to merge this PR short term?

I’m currently working on a follow-up PR that removes test.sql to reduce drift and the ongoing effort of maintaining two separate databases and want to avoid another stacked PR.

PR #697, opened 2026-01-16
security: XSS hardening across high-risk output surfaces

anonymoususer72041:
## Summary

This PR strengthens XSS protections across OpenCATS by making output escaping more consistent and safer in both legacy templates and dynamic UI output, especially where user-controlled or database-controlled content can reach HTML or JavaScript contexts.

It improves the core template escape helper (`Template::_`) (UTF-8, `ENT_QUOTES`, safer handling of null/arrays/non-stringable objects) and replaces a few direct `echo` outputs in error/questionnaire templates with the escaping helper to prevent raw HTML injection.

It also hardens additional high-risk areas: search result previews are now escaped while still highlighting matched keywords, several Career Portal placeholders and URLs are escaped/sanitized before insertion into HTML, JavaScript string injections are replaced with `json_encode(...)` for safe embedding, and Job Order `description`/`notes` are sanitized through a strict HTML allowlist (dropping script-like tags, unsafe attributes, and unsafe link protocols).

Finally, it adds a small PHPUnit unit test to validate that `Template::_` reliably escapes common script payloads and common entity-bypass patterns.

I did not add or adjust Behat scenarios here, so some end-to-end coverage might still be missing.

## Motivation

OpenCATS contains multiple rendering paths where content originating from users, imported resumes, database text fields, or configuration values can be inserted into HTML/JS without guaranteed context-appropriate escaping, which increases the risk of stored or reflected XSS (including in internal pages and the Careers Portal).

By centralizing and strengthening escaping behavior, sanitizing rich-text fields with a conservative allowlist, and ensuring JS/URL contexts are handled safely, this PR reduces the likelihood that a single unsafe field or template placeholder can become an executable payload while keeping the changes minimal and aligned with existing rendering patterns.


RussH:
I'll  review it shortly!

PR #704, opened 2026-01-28
fix: preserve job order regarding when editing

anonymoususer72041:
This PR fixes an issue where editing an Activity could unintentionally reset the "Regarding" field from a specific Job Order to "General". The root cause was fragile preselection logic in the edit UI that relied on comparing rendered text, which could fail (for example when the dropdown option text includes markers like `(*)`).

The edit modal now preselects the correct Job Order using the underlying `jobOrderID` rather than display text. In addition, saving "General" is normalized so that `activity.joborder_id` is persisted as SQL `NULL` instead of inconsistent sentinel values created by legacy paths (such as `0` or `-1`).

Finally, the PR includes a schema migration that updates existing rows and converts legacy `activity.joborder_id` values of `0` and `-1` to `NULL`, ensuring "General" is represented consistently across old and new data.

### DIFF
--- a/ajax/editActivity.php
+++ b/ajax/editActivity.php
@@ -69,7 +69,7 @@ $siteID = $interface->getSiteID();
 
 $activityID = $_POST['activityID'];
 $type       = $_POST['type'];
-$jobOrderID = $_POST['jobOrderID'];
+$jobOrderID = isset($_POST['jobOrderID']) ? trim($_POST['jobOrderID']) : null;
 
 /* Decode and trim the activity notes from the company. */
 $activityNote = trim(urldecode($_POST['notes']));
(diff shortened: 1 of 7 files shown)


RussH:
@anonymoususer72041 Can you fix this so it's not breaking the CI tests? Can't Merge until then of course.

anonymoususer72041:
@RussH I fixed the problem in `modules/install/Schema.php` and also adjusted missing changes in `modules/companies/Show.tpl`.

Unfortunately testing is currently not possible due to issue #735, so I will mark this PR as draft for now.

anonymoususer72041:
After #735 was solved, I was able to successfully complete the testing, so I'm marking this PR as ready to review again.

RussH:
Thank you for the PR - tested and merging.

PR #706, opened 2026-01-30
security: restrict AJAX during upgrade and escape installer config writes

anonymoususer72041:
## Summary

When upgrade mode is active and `INSTALL_BLOCK` is missing, `ajax.php` now only allows installer AJAX actions (`install:*`). All other AJAX requests are rejected early with the standard XML error response.

This PR also hardens installer config writes in `modules/install/ajax/ui.php` by escaping string values before passing them to `CATSUtility::changeConfigSetting()`. The affected installer settings now use safe PHP string literals via `var_export($value, true)` before being written to `config.php`.

## Motivation

During upgrades, `index.php` is already blocked when `INSTALL_BLOCK` is missing, but `ajax.php` remained accessible. That could allow existing sessions or stale UI tabs to continue sending AJAX requests while schema migrations are running, increasing the risk of unintended changes or inconsistent state.

Review also identified that several installer config values were still passed unescaped into `CATSUtility::changeConfigSetting()`, which then writes them directly into executable PHP in `config.php`. Escaping those values closes that remaining installer hardening gap without changing the intended installer flow.

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -106,6 +106,31 @@ if (!isset($_REQUEST['f']) || empty($_REQUEST['f']))
     die();
 }
 
+$installerActive = (!file_exists('INSTALL_BLOCK'));
+if ($installerActive)
+{
+    $module = '';
+    if (strpos($_REQUEST['f'], ':') !== false)
+    {
+        $parameters = explode(':', $_REQUEST['f']);
+        $module = preg_replace("/[^A-Za-z0-9]/", "", $parameters[0]);
+    }
+
+    if ($module !== 'install')
+    {
+        header('Content-type: text/xml');
+        echo '<?xml version="1.0" encoding="', AJAX_ENCODING, '"?>', "\n";
+        echo(
+            "<data>\n" .
+            "    <errorcode>-1</errorcode>\n" .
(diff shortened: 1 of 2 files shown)


RussH:
Thanks — definitely in the right direction, but checks show installer security is not fully hardened after this.

The remaining gap is that `modules/install/ajax/ui.php` still passes unescaped values into `CATSUtility::changeConfigSetting()`, which then writes them directly into executable PHP in `config.php`.

Could this PR also harden the config write paths by safely escaping the PHP string values before writing them, it could be as easy as using `var_export($value, true)` instead of concatenating raw input inside quotes before passing it to `changeConfigSetting()`, for;

* `DATABASE_USER`
* `DATABASE_PASS`
* `DATABASE_HOST`
* `DATABASE_NAME`
* `MAIL_SENDMAIL_PATH`
* `MAIL_SMTP_HOST`
* `MAIL_SMTP_USER`
* `MAIL_SMTP_PASS`
* `ANTIWORD_PATH`
* `PDFTOTEXT_PATH`
* `HTML2TEXT_PATH`
* `UNRTF_PATH`

That'd be ideal!

RussH (approved):

PR #708, opened 2026-02-02
fix: allow site admins to use administrative hide

anonymoususer72041:
This PR changes the permission threshold for the "Administrative Hide/Show" feature so that normal site administrators (ACCESS_LEVEL_SA) can use it, instead of restricting it to multi-site administrators (ACCESS_LEVEL_MULTI_SA).

Previously, site admins could neither toggle the hidden state nor view admin-hidden Candidates/Job Orders. With this change, site admins can access the action from the UI and can also open/view records that are marked as administratively hidden.


RussH (changes requested):
Thanks — I’ve reviewed the change itself and I’m happy with the direction here.

At this point the main thing blocking merge looks to be that the PR is currently marked as not mergeable, so could you please rebase this branch onto current master and resolve any conflicts? Should be good once that’s done!

RussH:
Thank you!

PR #710, opened 2026-02-04
security: add IP-based login throttling

anonymoususer72041:
This PR adds a simple brute-force protection for the OpenCATS login by rate-limiting failed login attempts per IP address.

If an IP address has more than 10 failed login attempts within the last 10 minutes, further login attempts from that IP are blocked early (before any credential validation) and a generic error message is shown.

To prevent bypassing the IP throttle by trying random usernames, failed attempts for unknown usernames are now also written to `user_login` (with `user_id = NULL` and `site_id = 0`).

### DIFF
--- a/lib/Session.php
+++ b/lib/Session.php
@@ -667,6 +667,50 @@ class CATSSession
     {
         $db = DatabaseConnection::getInstance();
 
+        if (isset($_SERVER['REMOTE_ADDR']))
+        {
+            $ip = $_SERVER['REMOTE_ADDR'];
+        }
+        else
+        {
+            $ip = '';
+        }
+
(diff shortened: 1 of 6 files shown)


RussH:
Thanks for this. I’m going to park this rather than merge backend login throttling into the legacy login flow.

For deployments with brute-force or exposure concerns, I think the better recommendation is to restrict staff/admin access at the web server, reverse proxy, firewall, or VPN layer. If an installation needs stronger protection, they can put the backend behind a VPN or IP allowlist and restrict unauthenticated access before requests reach OpenCATS.

If there is a specific authentication vulnerability, throttling would not be a sufficient mitigation anyway; we should fix that directly. I’d rather keep the backend login path simpler and focus current development effort on modernising the UI and adding practical features.

PR #713, opened 2026-02-12
refactor: stop persisting inline HTML for status change activity notes

anonymoususer72041:
## Summary

This PR stops persisting inline HTML markup for status change activity notes and cleans existing stored notes via a schema upgrade to remove previously persisted inline highlight markup.

## Motivation

Persisting presentation markup in the database mixes data and UI concerns, makes future styling changes harder and can cause inconsistent output across different views and exports. By storing plain text and removing legacy markup during upgrade, we keep activity notes clean and ensure consistent, maintainable rendering going forward.

### DIFF
--- a/ajax/editActivity.php
+++ b/ajax/editActivity.php
@@ -28,9 +28,7 @@
  */
 
 
-include_once(LEGACY_ROOT . '/lib/StringUtility.php');
 include_once(LEGACY_ROOT . '/lib/ActivityEntries.php');
-include_once(LEGACY_ROOT . '/lib/Pipelines.php');
 
 
 $interface = new SecureAJAXInterface();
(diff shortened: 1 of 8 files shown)


RussH (changes requested):
Thank you for the PR — this looks good (as always!), but on a more detailed review I found a couple of niggles I’d like fixed before merging:

The new highlighting helper is only applied in some activity views (candidates/Show.tpl, contacts/Show.tpl, activity/Search.tpl), but not all of them. For example, companies/Show.tpl still renders notes without the helper, so after the schema cleanup the orange status emphasis will disappear there.

There may be a double-escaping issue. In CandidatesUI.php the status change note is already passed through htmlspecialchars() before storage, and the new helper escapes again on output. That could display entities incorrectly for statuses containing special characters?

Please take a look and let me know.

anonymoususer72041:
I think we should mark this PR as draft for now.

I just found out that #702 is not working properly for Candidate Activities and still has problems with HTML entities, so I think we should fix this first before we continue work on this PR.

anonymoususer72041:
@RussH once #757 lands into master, I will rebuild this PR.

RussH:
@anonymoususer72041  #757 has been merged.

anonymoususer72041:
@RussH can you merge #761 first, please?

anonymoususer72041:
@RussH can you take a special look on 74c024669c507a4a9082a82873499bdbafca4857, please?

RussH:
Hi @anonymoususer72041 
Thanks — looking good, a few things though, 

1. Since master already has Template::escapeHtml(), could we use that inside TemplateUtility::highlightStatusChangeActivityNote() rather than calling htmlspecialchars() directly? it'd keep different parts of the codebase consistent.

For example:

    return Template::escapeHtml($notes);

and:

    return Template::escapeHtml($statusPrefix)
        . '<span class="statusChangeHighlight">'
        . Template::escapeHtml($statusText)
        . '</span>';

2. Can we make the schema migration tighter? At the moment it will clean any activity note containing a span with #ff6c00. Can we not  restrict that to status-change notes as well, e.g.:

    AND notes LIKE 'Status change: %'

That avoids altering unrelated historical notes which might contain similar markup.

PR #717, opened 2026-02-17
feat: sort activity lists newest-first

anonymoususer72041:
## Summary

This PR changes activity list ordering to show the newest entries first across the application where activity lists are rendered via the shared activity retrieval logic.

It updates the core activity queries to sort by the existing created-date sort field in descending order and adds a deterministic tie-breaker by `activity_id` (descending) to ensure stable ordering when timestamps are identical.

Additionally, the Activities module DataGrid ordering is hardened by appending an `activityID` tie-breaker when sorting by `dateCreatedSort`.

## Motivation

Seeing the most recent activity first is the expected UX for activity feeds and improves day-to-day usability when reviewing recent interactions.

The deterministic `activity_id` tie-breaker avoids unstable ordering caused by equal timestamps, making list rendering predictable and preventing confusing “random” reordering between page loads or environments.

### DIFF
--- a/lib/ActivityEntries.php
+++ b/lib/ActivityEntries.php
@@ -495,7 +495,8 @@ class ActivityEntries
             AND
                 activity.site_id = %s
             ORDER BY
-                dateCreatedSort DESC",
+                dateCreatedSort DESC,
+                activity.activity_id DESC",
             $this->_db->makeQueryInteger($dataItemID),
             $this->_db->makeQueryInteger($dataItemType),
             $this->_siteID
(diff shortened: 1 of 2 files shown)


anonymoususer72041:
Not sure if 691e59f is needed

RussH (changes requested):
Thanks, this looks good again!

One small point before merge: in the Activities datagrid path the tie-breaker is always appended as activityID DESC whenever sorting by dateCreatedSort. That makes ordering deterministic, but if a user explicitly sorts oldest-first the secondary ordering still runs newest ID first, which is a bit inconsistent with the chosen sort direction.

Could you either drop that second commit if it is not actually needed, or adjust it so the activityID tie-breaker follows the same direction as dateCreatedSort?

anonymoususer72041:
> Could you either drop that second commit if it is not actually needed, or adjust it so the activityID tie-breaker follows the same direction as dateCreatedSort?

It's actually not needed but useful, so I decided to provide a fix.

RussH (approved):
Thanks - this fix looks great.

PR #726, opened 2026-03-27
feat: rename Call activity to Not reached

anonymoususer72041:
## Summary

This PR renames the existing `Call` activity to `Not reached`.

The change updates the default schema for new installations, adjusts the relevant UI code paths that still used the `Call` label directly and adds a database migration so existing installations are updated as well.

The related activity types `Call (Talked)`, `Call (LVM)` and `Call (Missed)` remain unchanged.

## Motivation

The activity labels `Call` and `Call (Talked)` are too close to each other and can be confusing in daily use.

Using `Not reached` makes the distinction clearer, because it separates unsuccessful contact attempts from successful calls more explicitly.

This improves the activity selection and makes the intent of the existing activity type easier to understand without changing its underlying ID or behavior.

### DIFF
--- a/db/cats_schema.sql
+++ b/db/cats_schema.sql
@@ -70,7 +70,7 @@ CREATE TABLE `activity_type` (
 
 /*Data for the table `activity_type` */
 
-insert  into `activity_type`(`activity_type_id`,`short_description`) values (100,'Call');
+insert  into `activity_type`(`activity_type_id`,`short_description`) values (100,'Not reached');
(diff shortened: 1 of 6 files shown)


RussH (changes requested):
Thanks for the PR, looks good. But as the PR has a conflict in modules/install/Schema.php, could you please rebase this onto current master and resolve that conflict, I’ll be happy to merge after that.

anonymoususer72041:
I think we just crossed paths. Will merge this now.

PR #729, opened 2026-03-30
fix: correct candidate re-apply update mapping and guard empty ownership emails

anonymoususer72041:
Fixes #725

This PR corrects the misaligned `Candidates::update()` call in the careers re-apply flow.

It also adds a small safeguard in `lib/Candidates.php` so candidate ownership notifications are only sent when both the recipient and email body are non-empty.


RussH:
Good to merge

PR #527, opened 2021-02-21
Merging develop back to master. Develop has been installed locally for UAT (passed)

RussH:
Merging develop back to master (new changes: travis-ci fix, mailer.php fix, iterate version number). Develop has been installed locally for UAT (passed)

PR #493, opened 2020-09-11
Merge develop back to Master for a new release. 

RussH:
This will be the basis for a PHP 7.2 compatible release from Master.

PR #734, opened 2026-03-31
chore: merge develop into master

RussH:


anonymoususer72041:
I did a thorough analysis of the develop → master delta and would recommend against a full merge. Here's why:

develop has 127 commits not in master, but master has diverged significantly with 164 commits of its own. The merge-base shows that most of develop's changes were already incorporated into master via earlier merges (#493, #527).

I rebased the 127 develop commits onto upstream/master as a test. After resolving all conflicts, only 14 files remained with actual diffs — and most of them are harmful:

- get_magic_quotes_gpc() / get_magic_quotes_runtime() — reintroduced in index.php, QueueCLI.php, lib/Attachments.php, lib/InstallationTests.php, modules/import/ImportUI.php. These functions were removed in PHP 8.0 and will crash at runtime. Master already removed them intentionally.
- config/sphinx/sphinx.conf — ended up with unresolved merge conflict markers.
- lib/sphinx/conf/sphinx.conf — contains hardcoded database credentials.
- config.php — duplicate commented-out encoding block.

Out of 683 changed files, only 2 genuinely valuable fixes are missing from master:

- lib/Mailer.php:208 — unescaped double quote in an HTML href attribute (http://www.opencats.org" should be http://www.opencats.org\")
- js/lib.js — missing var DATA_ITEM_DUPLICATE = 900; (the PHP-side constant already exists in constants.php, but the JS counterpart is absent)

Suggestion: Instead of merging develop, cherry-pick or manually apply just these two one-line fixes. Then consider retiring the develop branch by resetting it to master to avoid future confusion.

Issue #735, opened 2026-03-31
Editing activities fails with "Invalid request"


anonymoususer72041:
### Expected behavior and actual behavior

Activities should be editable from the UI without a server-side request error. Currently, attempting to edit and save an activity fails with an "Invalid request" error message.

### Steps to reproduce the problem

1. Open any existing activity.
2. Edit the activity.
3. Save the changes.
4. Observe that the request fails with an "Invalid request" error message.

### What version of OpenCATS are you running?

Current master branch.

#### Release or downloaded from Git?

Downloaded from Git / current master branch.

#### WAMP or LAMP?

Docker test environment

### What version of PHP and MySQL are you running

Docker test environment

anonymoususer72041:
I can no longer reproduce this issue on the current master branch after forcing a hard reload and bypassing cached JavaScript assets in the browser.

The observed behavior was caused by stale client-side assets, so the browser was still executing an outdated js/lib.js version. After reloading with cache disabled, the affected AJAX requests worked again.

Closing as not reproducible on current master.

PR #749, opened 2026-04-09
fix: add file-based cache busting for JavaScript and stylesheet assets

anonymoususer72041:
This PR replaces the previous legacy asset anti-cache handling for JavaScript and stylesheet files with centralized file-based cache busting.

The change was motivated by stale browser-cached assets causing outdated client-side code to run against newer server-side code. In practice, this can produce misleading regressions that appear to be application bugs but disappear after a forced reload.

During investigation, issues such as #735, #739 and #740 were found to be affected by this class of problem. In those cases, stale JavaScript assets could cause AJAX requests to behave differently from the current server-side code, which made current master appear broken until the browser cache was bypassed. This PR addresses that underlying asset versioning problem so updated JavaScript and stylesheet files are fetched reliably after changes.

### DIFF
--- a/installtest.php
+++ b/installtest.php
@@ -33,6 +33,24 @@ include_once('./config.php');
 include_once(LEGACY_ROOT . '/constants.php');
 include_once(LEGACY_ROOT . '/lib/InstallationTests.php');
 
+/* Version check before including TemplateUtility for asset URL versioning. */
+$phpVersion = phpversion();
+$phpVersionParts = explode('.', $phpVersion);
+if ($phpVersionParts[0] >= 5)
+{
+    include_once(LEGACY_ROOT . '/lib/TemplateUtility.php');
+}
+else
+{
+    $php4 = true;
+}
+
+$mainCSSURL = 'main.css';
+if (!isset($php4))
+{
+    $mainCSSURL = call_user_func(array('TemplateUtility', 'getVersionedAssetURL'), $mainCSSURL);
+}
+
 
 define('REQUIRED_SCHEMA_VERSION', '1200');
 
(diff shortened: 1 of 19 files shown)


anonymoususer72041:
I was not able to test this change in a fully reliable way before opening the PR.

Additional testing would be appreciated, especially around shared page headers, direct asset includes and pages with non-standard layouts.

RussH:
I'll check this out locally and test

RussH:
Thanks — the asset cache-busting change looks useful, and I do not see a PHP 7.2 compatibility issue in that part.

The main thing I’d ask for before merge is to make the activity notes rendering consistent. These two files are using the safer escaped pattern:

* `modules/companies/Show.tpl`
* `modules/contacts/Show.tpl`

They now do:

```php id="n8ortr"
nl2br(htmlspecialchars($activityData['notes'], ENT_QUOTES | ENT_SUBSTITUTE, HTML_ENCODING))
```

That is ideal, the note text is escaped first and only then converted to `<br>` for display.

But these other paths are still using `nl2br()` on raw notes without escaping first:

* `ajax/getPipelineDetails.php`
* `modules/activity/Search.tpl`
* `modules/activity/dataGrids.php`
* `modules/candidates/Show.tpl`

So at the moment the PR leaves note output handled in two different ways, and these files are still  vulnerable to XSS exploit if `notes` contains HTML or script content.

Could those be aligned so all note rendering uses the same escaped pattern before `nl2br()`?

Finally -  this PR seems broader than the title suggests, since it also includes the `Status Change` activity type/schema change and some unrelated cleanup. Splitting those out would make the cache-busting change much easier to review and would keep the PR's aligned.

anonymoususer72041:
I think you mixed things up with #733, #737 and #745, @RussH.

Regarding activity notes, I will create a new PR.

RussH:
Thanks — I’ve rechecked this and I think my earlier comment was mixed up with other PRs. Looking at the current diff, I don’t see any issues.

PR #751, opened 2026-04-15
security: escape activity notes before applying nl2br

anonymoususer72041:
This PR addresses @RussH's review feedback on PR #749 regarding inconsistent activity note rendering.

Some activity note outputs already escaped note content before applying `nl2br()`, while several remaining paths still rendered raw note content. This inconsistency left those locations exposed to an XSS risk if notes contained HTML or script content.

The remaining note rendering paths are now aligned with the safer pattern already used elsewhere in the codebase: `nl2br(htmlspecialchars($notes, ENT_QUOTES | ENT_SUBSTITUTE, HTML_ENCODING))`

### DIFF
--- a/ajax/getPipelineDetails.php
+++ b/ajax/getPipelineDetails.php
@@ -77,7 +77,7 @@ else
              $activity['enteredByLastName'],
              ')</td>';
         echo '<td style="padding-right: 6px; width: 625px;">',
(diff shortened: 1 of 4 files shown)


RussH (approved):
Nice small fix,  it brings the remaining activity note render paths into line with the safer escaped pattern already used elsewhere.

PR #755, opened 2026-04-22
refactor: migrate bundled lib dependencies to Composer

anonymoususer72041:
This PR replaces bundled third-party libraries under `lib/` with Composer-managed dependencies where this could be done without broader functional changes.

The bundled FPDF copy was replaced with `setasign/fpdf` and the bundled Sphinx PHP API was replaced with `neutron/sphinxsearch-api`. Related runtime includes and paths were updated accordingly.

`lib/artichow` is intentionally not part of this migration. It is still actively used by the current graph generation code and does not have a straightforward Composer-based replacement in this context.

The Sphinx client package used here is unmaintained, so this should be understood as a dependency management improvement, not as a long-term modernization of the Sphinx integration.


RussH (changes requested):
Good cleanup direction here — replacing the old  FPDF/Sphinx files with Composer-managed dependencies makes sense, and removing FPDF code is food - going towards  PHP 8.x.

A few things I think need fixing before merge:

- setasign/fpdf now requires ext-gd and ext-zlib, so we need to confirm/update the PHP/CI environment so composer install does not fail where GD is missing.
- ReportsUI.php no longer explicitly includes FPDF. That is fine only if vendor/autoload.php is guaranteed to be loaded globally before generateJobOrderReportPDF() runs; otherwise this will fatal with Class 'FPDF' not found.
- config/sphinx/sphinx.conf now hard-codes /usr/local/www/catsone.com/data/config/sphinx/STOPWORDS, which makes the config non-portable for normal installs/dev/CI. This probably needs to be changed, documented, or left as a path the installer/admin updates.

Can you take a look and see what clean-up can be done  merging?

anonymoususer72041:
> * setasign/fpdf now requires ext-gd and ext-zlib, so we need to confirm/update the PHP/CI environment so composer install does not fail where GD is missing.

I will push a tiny fix for that.

I wonder how we want to deal with legacy travis CI? Is there any specific reason why travis is still in this repo, @RussH?

> * ReportsUI.php no longer explicitly includes FPDF. That is fine only if vendor/autoload.php is guaranteed to be loaded globally before generateJobOrderReportPDF() runs; otherwise this will fatal with Class 'FPDF' not found.

I mixed this up with another unpublished branch of mine where I change OpenCATS to PSR-4. I will fix this in this PR.

RussH:
> I wonder how we want to deal with legacy travis CI? Is there any specific reason why travis is still in this repo, @RussH?

No need for Travis CI in this repo any longer, we can pull that

RussH (approved):
There may still be some follow-up work needed around the optional Sphinx configuration/scripts? The Sphinx config has moved from `lib/sphinx/conf/sphinx.conf` to `config/sphinx/sphinx.conf`,  but may still contain old `lib/sphinx/...` references. I don’t think that needs to block this cleanup PR, since Sphinx is optional/disabled by default. 

Happy to merge this as-is though, and handle any Sphinx-specific cleanup separately once we have a cleaner codebase.

PR #758, opened 2026-04-23
feat: allow manually setting activity date and time

anonymoususer72041:
## Summary

This PR allows users to manually set the date and time when logging an activity.

It adds manual date and time inputs to the activity modal and persists the selected timestamp when the activity is created. If no valid manual timestamp is provided, the existing default behavior is preserved.

The change is intentionally minimal and keeps the existing flow intact while making sure manually logged activities can reflect their actual occurrence time.

## Motivation

When activities are entered after the fact, using the current timestamp may not accurately reflect when the activity originally occurred.

This is especially relevant for emails that are logged retroactively. In those cases, it is helpful for the recorded activity to reflect the original date and time instead of the time of entry.

### DIFF
--- a/lib/ActivityEntries.php
+++ b/lib/ActivityEntries.php
@@ -82,14 +82,25 @@ class ActivityEntries
      * @param string Activity notes.
      * @param integer Entered-by user ID.
      * @param integer Job Order ID; -1 for general (stored as NULL).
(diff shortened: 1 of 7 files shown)


anonymoususer72041:
Tested and verified in production. Merging now.

PR #761, opened 2026-04-26
security: harden template output escaping

anonymoususer72041:
This PR centralizes template output escaping helpers and applies them to selected OpenCATS UI outputs.

It adds reusable escaping methods for HTML text, HTML attributes, URLs, and JavaScript string literals, then updates existing template output to use the appropriate helper for each context. The goal is to make escaping behavior more consistent and reduce the risk of XSS issues while keeping the changes minimal and close to the existing template structure.

The updated areas include template header output, candidate detail page output, contact detail page output, company detail page output, job order detail page output and activity-related output.

The changes intentionally avoid business logic changes, SQL changes, permission changes and layout restructuring. Activity notes continue to preserve line breaks where applicable, but are escaped at output time before formatting.


RussH (approved):
This looks clean - so long as there's no html expected in activities or questionnaire responses.  The tests are good, too - thank you!

PR #762, opened 2026-05-01
refactor: move activity description escaping to templates

anonymoususer72041:
This PR moves activity description escaping for activity schedule event modals from the controller layer to the template output layer.

Previously, the contact and candidate activity modal descriptions were escaped before being assigned to the template, while the corresponding templates rendered the assigned value directly. This worked, but mixed presentation escaping into the controller flow and made it easier to accidentally introduce inconsistent escaping behavior.

The controllers now assign the raw activity note to `activityDescription`, and the modal templates escape the value at the output boundary using the existing OpenCATS template escaping helper. This keeps the value escaped exactly once, avoids double escaping and aligns the contact and candidate activity schedule event modals with the preferred template output pattern.

The change is intentionally limited to the contact and candidate activity schedule event modal description flow. It does not change activity creation, request handling, database handling, calendar event behavior or modal behavior.

### DIFF
--- a/modules/candidates/AddActivityScheduleEventModal.tpl
+++ b/modules/candidates/AddActivityScheduleEventModal.tpl
@@ -198,7 +198,7 @@
          <?php if (!$this->onlyScheduleEvent): ?>
             <?php if ($this->activityAdded): ?>
                 <?php if (!empty($this->activityDescription)): ?>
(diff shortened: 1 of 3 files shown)


RussH:
Looks good, passes tests. Thank you!

PR #763, opened 2026-05-01
fix: add CSRF token output to install wizard

anonymoususer72041:
When the install wizard is opened while an authenticated OpenCATS session is still active, its initial AJAX request can fail with "-1 Invalid request." because ajax.php enforces CSRF validation for logged-in sessions, but installwizard.php did not expose CATSCsrfToken for js/lib.js to append to the request.

This change starts the existing OpenCATS session before any HTML output in installwizard.php, loads the required session and template classes and outputs CATSCsrfToken when an authenticated session is present.

The CSRF validation in ajax.php remains unchanged. Logged-out install wizard access continues to behave as before, while logged-in sessions now provide the token needed for installer AJAX requests.

### DIFF
--- a/installwizard.php
+++ b/installwizard.php
@@ -1,19 +1,18 @@
-<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
-"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <?php
     include_once('constants.php');
     include_once('config.php');
 
-    /* We aren't using any TemplateUtility methods that require us to pull in
-     * any of its dependencies.
-     */
     /* Version check before we include this. */
-   
     $phpVersion = phpversion();
     $phpVersionParts = explode('.', $phpVersion);
     if ($phpVersionParts[0] >= 5)
     {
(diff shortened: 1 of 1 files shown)


RussH:
@anonymoususer72041 
One thing I’d like checked before merge: does Template::escapeJs() return a complete quoted JavaScript string literal, or only an escaped string?

At the moment this renders the token directly into JavaScript:

echo '<script type="text/javascript">CATSCsrfToken = ',
     Template::escapeJs($_SESSION['CATS']->getCSRFToken()), ';</script>', "\n";

If escapeJs() does not add the surrounding quotes itself, this should be changed to emit a proper JS string, -  wrap the escaped value in quotes?

anonymoususer72041:
@RussH I checked this against the current implementation of Template::escapeJs(). It returns a complete quoted JavaScript string literal, not just an escaped string.

The method uses json_encode() with JSON_HEX_TAG, JSON_HEX_AMP, JSON_HEX_APOS and JSON_HEX_QUOT, and returns the encoded value directly. For example, Template::escapeJs('safe value') returns "safe value", including the surrounding quotes.

So the current installwizard.php output is already emitting a valid JavaScript string literal here and should not need additional quotes around the escaped value.

PR #764, opened 2026-05-02
feat: exclude closed jobs from activity references

### LINKED ISSUE #361: Logging an activity should only offer open jobs
### Expected behavior and actual behavior.
When logging an activity against a client contact, it should only offer live jobs and not closed ones. 

### Steps to reproduce the problem.
open a client contact who has both live and closed vacancies. The drop-down will offer both live and closed vacancies.
### What version of opencats are you running? WAMP or LAMP? 
N/A
### attach appropriate error logs. Please attach [apache/mysql] error/access logs as needed.
N/A

anonymoususer72041:
## Summary

This PR updates the add activity / schedule event flows for Contacts and Candidates so that the Regarding job reference dropdown excludes job orders from the configured Closed status group.

Instead of only allowing jobs from the Open status group, the filtering now excludes Closed-group statuses. This keeps Pre-Open job orders such as Upcoming and Lead available, while hiding Closed, Canceled or other statuses configured as closed.

The change adds a central closed job order status SQL helper and uses it from the Contact and Candidate activity reference queries.

The existing activity edit AJAX flow is intentionally left unchanged.

## Motivation

When creating new activities, the Regarding dropdown should focus on current job orders that are still relevant for new activity logging. Closed job orders are no longer current references in that context and can make the selection unnecessarily noisy.

This is also related to #361.

### DIFF
--- a/lib/Contacts.php
+++ b/lib/Contacts.php
@@ -33,6 +33,7 @@ include_once(LEGACY_ROOT . '/lib/Pager.php');
 include_once(LEGACY_ROOT . '/lib/EmailTemplates.php');
 include_once(LEGACY_ROOT . '/lib/ExtraFields.php');
 include_once(LEGACY_ROOT . '/lib/Calendar.php');
+include_once(LEGACY_ROOT . '/lib/JobOrderStatuses.php');
 
 /**
  *	Contacts Library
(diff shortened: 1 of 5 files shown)


RussH:
Looks good to merge

PR #766, opened 2026-05-02
fix: prefill installer phone country code from site settings

anonymoususer72041:
This PR updates the install wizard so the default phone country calling code field is prefilled from the existing `site.default_phone_country_code` value when available.

Previously, the install wizard always rendered the local phone country calling code input as empty, even if a value was already stored in the database. This could cause the existing setting to be overlooked during installer runs.


anonymoususer72041:
In this exceptional case, it makes sense to wait to merge this fix until version 0.9.8.0 is released.

The Default Phone Country Code is new and not yet included in any installation. Therefore, when upgrading with the introduction of the Default Phone Country Code feature, users will have to explicitly select a Default Phone Country Code and it won't be pre-assigned with +1 and potentially overlooked.

Of course, it wouldn't be a problem if this fix were included in version 0.9.8.0 anyway.

RussH:
Lets hold this until the next release, we'll get the current release out the door first

RussH:
Looks good to comit, will update the branch after pre-req PR #742  is merged.

RussH (approved):

PR #767, opened 2026-05-02
chore!: remove legacy database upgrade path

anonymoususer72041:
This PR removes the obsolete legacy database upgrade path that depended on standalone `db/upgrade-*.sql` files.

The removed installer path previously detected historical OpenCATS database revisions and applied old version-to-version SQL upgrade scripts.

Fresh installations continue to use `db/cats_schema.sql`. Existing compatible OpenCATS installations continue through the current installer flow and schema maintenance path.

The optional ZIP code upgrade file is intentionally kept because it is still used by the optional ZIP code component.


RussH (changes requested):
Thanks, good cleanup.

One point to clarify before merge: the PR description says the optional ZIP code upgrade file is intentionally kept because it is still used by the optional ZIP code component. My understanding is that the older ziplookup code has effectively been superseded by lib/ZipLookup.php, and this PR also removes the upgradeCats path which previously ran upgrade-zipcodes.sql. Could you please confirm whether that ZIP upgrade file is still genuinely used by current code? If not, I’d prefer either removing it in this PR or noting it as a follow-up cleanup.

Assuming that ZIP lookup point is clarified, this looks good to me.

anonymoususer72041:
I checked this again and `db/upgrade-zipcodes.sql` is still used by current code outside of the removed legacy installer upgrade path.

The removed `upgradeCats` path was one place where it could previously be executed, but the optional `usZipCodes` component still installs its data from `db/upgrade-zipcodes.sql` in `modules/install/OptionalComponents.php`. Install schema revision `253` in `modules/install/Schema.php` also still uses the same file when existing ZIP code data is detected.

The runtime ZIP distance filtering is still tied to the local `zipcodes` table when `US_ZIPS_ENABLED` is enabled, so removing this file in this PR would break the current optional ZIP code component path.

I agree that the filename is a bit misleading now that the legacy upgrade path is gone. Renaming or moving it could make sense as a separate follow-up cleanup, but I think it should stay in this PR.

PR #768, opened 2026-05-04
fix: use NULL for missing calendar event job order references

anonymoususer72041:
PR #704 changed optional activity job order references to use NULL instead of the legacy -1 sentinel. Calendar events have the same optional job order relationship, but `calendar_event.joborder_id` still used `-1` for missing job order references.

This updates `calendar_event.joborder_id` so missing job order references are represented as NULL consistently. The base schema now allows NULL and defaults to NULL and the update migration converts existing `-1` values as well as orphaned non-NULL job order references to NULL.

The runtime write paths were updated accordingly. Calendar event creation and updates now pass PHP null when no job order is associated, and `Calendar::addEvent()` / `Calendar::updateEvent()` write SQL NULL for that case. Deleting a job order now clears related calendar event job order references by setting them to NULL.

The scope is intentionally limited to `calendar_event.joborder_id`. Other legacy `-1` sentinels, such as `calendar_event.data_item_id`, `calendar_event.data_item_type`, activity input sentinels, and filter parameters where `-1` has a different meaning, are left unchanged.

### DIFF
--- a/db/cats_schema.sql
+++ b/db/cats_schema.sql
@@ -123,7 +123,7 @@ CREATE TABLE `calendar_event` (
   `date_created` datetime NOT NULL DEFAULT '1000-01-01 00:00:00',
   `date_modified` datetime NOT NULL DEFAULT '1000-01-01 00:00:00',
   `site_id` int(11) NOT NULL DEFAULT '0',
-  `joborder_id` int(11) NOT NULL DEFAULT '-1',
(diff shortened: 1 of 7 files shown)


RussH:
Suggest we pin this for the next release, however;

Only thing to check before merge is that modules/install/Schema.php still applies cleanly against current master, as that file often picks up conflicts.  Happy to merge.

PR #771, opened 2026-05-05
fix: replace deprecated mysql_* with mysqli_* in upgrade scripts

### LINKED ISSUE #572: Demo content doesn't work

Frankli9986:
## Problem
Fixes #572

When installing OpenCATS with demo content on PHP 7+, the installation fails because database upgrade scripts still use the deprecated `mysql_*` functions, which were removed in PHP 7.0.

## Changes
- `mysql_real_escape_string()` → `mysqli_real_escape_string()` (2 occurrences)
- `mysql_fetch_row()` → `mysqli_fetch_row()` (1 occurrence)
- Added `global \;` where needed to ensure the mysqli connection resource is available in the upgrade script context.

## Files Modified
- `modules/install/Schema.php`

## Testing
- Verified that no other `mysql_*` function calls remain in the codebase (only these 3 were found in upgrade script strings).
- The fix aligns with the existing `mysqli_*` usage already present in `modules/install/ajax/ui.php`.

## Related Issue
- #572

### DIFF
--- a/modules/install/Schema.php
+++ b/modules/install/Schema.php
@@ -700,6 +700,7 @@ class CATSSchema
              ',
              /* Convert dashboard component parameters to serialize()d arrays. */
              '225' => 'PHP:
+                global $mySQLConnection;
                 include_once(\'./lib/ListEditor.php\');
 
                 $rs = $db->getAllAssoc(
@@ -722,7 +723,7 @@ class CATSSchema
                         "UPDATE
                             dashboard_component
                         SET
-                            module_parameters = \'" . mysql_real_escape_string($serializedValue) . "\'
+                            module_parameters = \'" . mysqli_real_escape_string($mySQLConnection, $serializedValue) . "\'
                         WHERE
                             dashboard_component_id = " . $row[\'dashboard_component_id\']
                     );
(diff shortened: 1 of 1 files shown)


anonymoususer72041:
Hi @Frankli9986, thanks for looking into this. This looks closely related to and partly overlapping with my existing PR #765.

From what I can see, this PR fixes the same obsolete `mysql_*` calls in `modules/install/Schema.php` that are already covered in #765. The main difference is that #771 replaces them with direct `mysqli_*` calls, while #765 routes them through the existing `DatabaseConnection` API where possible. In particular, #765 also touches the same install schema areas and includes a broader cleanup of legacy database result handling around install, backup and control panel code.

I think it would be better to consolidate the work in one PR rather than having two separate fixes for the same issue. You are very welcome to join #765 and submit suggestions there, especially because there are small differences between both approaches that may still be useful to discuss or incorporate.

I also started a discussion in the comments of #765 about a few open questions before merging, including whether changing previous migrations is the right approach. I would be happy if you joined that discussion as well.

Frankli9986:
Hi @anonymoususer72041, thanks for the thoughtful pointer to #765 and apologies for the late reply.

You're right that #765 already covers the `modules/install/Schema.php` changes in this PR, with the broader benefit of routing through `DatabaseConnection` rather than calling `mysqli_*` directly — which aligns better with the codebase's direction. Given that #765 is actively progressing and has @RussH's agreement on the short-term compatibility approach, it makes more sense to consolidate there.

I'll close this PR shortly. Happy to help test the demo-content install path on PHP 7/8 against your branch once it's ready for a final review — that specific scenario (#572) is what brought me here originally, so I can confirm it works end-to-end.

Thanks again for the careful review and for keeping the cleanup coherent.

PR #776, opened 2026-05-05
refactor: run security visibility tests without JavaScript

anonymoususer72041:
This PR removes unnecessary `@javascript` tags from security visibility Behat scenarios that only verify server-rendered page content and access-level visibility.

The `@javascript` tag is intentionally kept for `Calendar module visibility` and `Job Order Show page visibility`, because validation showed that these scenario outlines still require the JavaScript/Selenium session.

This keeps the security visibility coverage unchanged while allowing most of these scenarios to run through the faster non-JavaScript Behat session.


anonymoususer72041:
Seems like CI testing is already ~2 minutes faster

PR #782, opened 2026-05-08
refactor: share add activity schedule event modal

anonymoususer72041:
This refactors the duplicated add activity / schedule event modal templates for candidates and contacts into a shared activity template.

The candidate, contact and job order entry points now pass the required modal configuration through template variables, while preserving the existing differences between those flows. This includes the parent module/action, hidden parent ID field, regarding job order options, modal title, validator path, cancel and close URLs, event duration visibility, field widths, reminder label text and focus behavior.

The candidate and contact regarding lists are normalized before rendering so the shared template can use the same structure for both flows. Candidate labels continue to include the related company name where applicable, while contact labels remain title-only.

A follow-up fix also removes undefined variable usage in the candidate job order finished path by avoiding access to unavailable pipeline data and using the current regarding job order ID for the cancel URL.


anonymoususer72041:
> The candidate, contact and job order entry points now pass the required modal configuration through template variables, while preserving the existing differences between those flows. This includes the parent module/action, hidden parent ID field, regarding job order options, modal title, validator path, cancel and close URLs, event duration visibility, field widths, reminder label text and focus behavior.

We might want to check if we should align some differences here before merging.

PR #787, opened 2026-05-14
chore: upgrade to PHP 7.4

anonymoususer72041:
This PR updates OpenCATS to use PHP 7.4 as the target runtime. It raises the Composer PHP requirement to PHP 7.4, updates the Composer dependencies and lock file to PHP 7.4-compatible versions and updates the CI configuration from PHP 7.2 to PHP 7.4.

The Docker test setup now uses the local PHP container for Composer instead of running Composer through a separate generic Composer image with ignored platform requirements. This makes the documented local testing flow closer to the actual PHP runtime used by the test containers and avoids resolving dependencies against a mismatching PHP environment.

The codebase is also adjusted for PHP 7.4 compatibility by removing deprecated magic quotes checks, replacing deprecated curly brace string offsets, replacing deprecated `create_function()` usage, replacing deprecated `mcrypt` usage with OpenSSL and fixing deprecated `implode()` argument order. Some of these changes may become unnecessary if #778 is merged first, as that PR removes several unused legacy helpers affected by the same PHP 7.4 compatibility issues.

These changes keep the PHP 7.4 upgrade focused on compatibility, dependency updates and the local/CI test setup without introducing unrelated behavior changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: ['7.2']
+        php-version: ['7.4']
 
     steps:
       - name: Checkout Code
--- a/QueueCLI.php
+++ b/QueueCLI.php
@@ -55,20 +55,6 @@ include_once(LEGACY_ROOT . '/modules/queue/constants.php');
 @session_name(CATS_SESSION_NAME);
 session_start();
 
-/* Make sure we aren't getting screwed over by magic quotes. */
-if (get_magic_quotes_runtime())
-{
-    if (function_exists('set_magic_quotes_runtime')) {
-        set_magic_quotes_runtime(0);
-    }
-}
-if (get_magic_quotes_gpc())
-{
-    $_GET     = array_map('stripslashes', $_GET);
-    $_POST    = array_map('stripslashes', $_POST);
-    $_REQUEST = array_map('stripslashes', $_REQUEST);
-}
-
 if (!isset($_SESSION['CATS']) || empty($_SESSION['CATS']))
 {
     $_SESSION['CATS'] = new CATSSession();
--- a/README-testing.md
+++ b/README-testing.md
@@ -26,18 +26,18 @@ To mirror the CI environment on your machine:
 2. **Start the containers:**
    ```bash
    cd docker/
-   docker compose -f docker-compose-test.yml up -d
+   docker compose -f docker-compose-test.yml up -d --build
    ```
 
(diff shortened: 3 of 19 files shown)


anonymoususer72041:
~Note: This PR is currently stacked on top of #756, so that PR should be reviewed and merged first.~

RussH (changes requested):
Good PR for the PHP 7.4 uplift. 

The one thing I’d like tightened before merge is the newly added Docker build process. It downloads  unrtf over http,  unrtf should at least move from the old HTTP 0.21.9 URL to the current GNU HTTPS release, ideally https://ftp.gnu.org/gnu/unrtf/unrtf-0.21.10.tar.gz.

Also worth confirming whether "php": "^7.4" is intentional, as that restricts installs to PHP 7.4 only. If we want this as a stepping stone toward PHP 8.x, >=7.4 or ^7.4 || ^8.0 may be more suitable.

If you can update the unrtf target to 0.21.10 and transport to https  I'll merge afterwards.

anonymoususer72041:
> The one thing I’d like tightened before merge is the newly added Docker build process. It downloads unrtf over http, unrtf should at least move from the old HTTP 0.21.9 URL to the current GNU HTTPS release, ideally https://ftp.gnu.org/gnu/unrtf/unrtf-0.21.10.tar.gz.

Good catch, thanks. I validated this and the Dockerfile was indeed downloading unrtf 0.21.9 over HTTP from the old GNU URL. I’ll update it to use the current GNU HTTPS release at https://ftp.gnu.org/gnu/unrtf/unrtf-0.21.10.tar.gz.

> Also worth confirming whether "php": "^7.4" is intentional, as that restricts installs to PHP 7.4 only. If we want this as a stepping stone toward PHP 8.x, >=7.4 or ^7.4 || ^8.0 may be more suitable.

I checked this as well. `^7.4` is intentional here, at least for this PR.

I ran a temporary PHP 8.0 smoke test to see whether broadening the constraint would already be safe. The PHP 8.0 Docker image builds, Composer install succeeds after temporarily relaxing the constraint, `composer check-platform-reqs` passes and the UnitTests pass. However, that still does not look like enough evidence to advertise PHP 8 support yet: the run shows PHP 8 deprecation warnings, and the IntegrationTests/Behat runs are not clean yet due to bootstrap/environment issues that need to be investigated separately.

Because of that, I would prefer to keep `"php": "^7.4"` in this PR. This change raises the supported baseline from PHP 7.2 to PHP 7.4 only. PHP 8 compatibility should be handled in a separate follow-up PR once the remaining compatibility and test-suite issues have been addressed.

PR #795, opened 2026-06-03
security: deny Apache web access to temp directory

anonymoususer72041:
Deny direct HTTP access to the temp directory for Apache deployments by adding a dedicated .htaccess file.

The root .htaccess disables directory indexes and sets security headers, but it does not prevent direct access to known files below temp/. This change mirrors the existing deny-by-default approach used for attachments/ and helps prevent temporary files from being served directly by Apache when AllowOverride is enabled.

The temp/empty placeholder is removed because the new versioned temp/.htaccess file now keeps the temp directory present in the repository. The .gitignore rules are updated so temporary files remain ignored while temp/.htaccess stays tracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ docker/persist/*
 vendor/*
 attachments/*
 temp/*
+!temp/.htaccess
 uploads/*
 # Ignore Mac DS_Store files
 .DS_Store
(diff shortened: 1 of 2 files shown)


RussH (approved):
Good security hardening.

PR #797, opened 2026-06-04
fix: prevent PHP errors in direct AJAX endpoints

anonymoususer72041:
This PR adds a shared AJAX bootstrap file and uses it from AJAX endpoints that may be accessed directly. Previously, some direct AJAX requests could fail because required constants or classes were not loaded before endpoint-specific code executed.

The change defines the legacy root path consistently for AJAX requests, loads the required core files, and updates affected includes to use root-based paths where needed. This avoids errors such as missing `LEGACY_ROOT`, missing AJAX interface classes and failed relative includes when endpoints are executed from a different working directory.

Validation was performed by linting the changed PHP files and directly executing the previously problematic template AJAX endpoints. The endpoints now return the expected not-logged-in XML response instead of producing include or bootstrap-related PHP warnings.


ocjorge:
I tested a similar fix in a production-like environment (Debian 12, PHP 8.4.21). 
The LEGACY_ROOT bootstrap issue also affected rss/index.php which I patched 
separately in my now-closed PR #809. This approach of a shared bootstrap 
file is cleaner and more maintainable than individual per-file defines. LGTM.

RussH (changes requested):
Can you check  that ajax/bootstrap.php is included first in each direct AJAX endpoint before any legacy library includes, doesn't seem to be first in every instance.

anonymoususer72041:
Thanks, I checked the changed direct AJAX endpoints again. In this branch, `ajax/bootstrap.php` is included before the legacy library includes in the affected endpoints.

While checking this, I also noticed that `modules/install/ajax/ui.php` still contains some pre-existing relative includes such as `include_once('lib/SystemUtility.php')` in the resume parsing path. Those lines were already present before this branch and were not introduced by this commit. Since they are a separate legacy relative-path issue rather than part of the bootstrap-order change, I would prefer to handle them in a follow-up PR to keep this fix focused.

RussH (approved):

PR #802, opened 2026-06-04
refactor: remove legacy licensing

anonymoususer72041:
This PR removes the obsolete legacy licensing system from OpenCATS.

It removes runtime license key validation, the old license key configuration entries, legacy license wizard steps, Professional Membership settings, professional feature gates and the remaining legacy license utility. OpenCATS no longer attempts to validate, register, refresh, persist or depend on a license key at runtime.

The candidate resume handling was kept intact while removing the professional feature gates. In particular, pasted resume text submitted through `textResumeBlock` is still handled and can still create a text resume attachment.


RussH (commented):
Thanks — this looks good to me.

The only thing I’d like confirmed before merge is the runtime behaviour around resume parsing now that the licence requirement has been removed. I’m not concerned about preserving the current external SOAP parser long-term, as that will likely be replaced, but can you confirm that if the parser service is unavailable, the UI does not hang while waiting for a network timeout and the user gets a clean failure/fallback?

Assuming that is confirmed, I’m happy to merge this.

anonymoususer72041:
Thanks, I checked this and found that the current branch should not be treated as fully confirmed yet.

The parser failure path already returns `false` for SOAP faults and the callers mostly fall back to the unparsed resume text / empty parsed fields, so a parser-side failure does not need to block candidate creation or resume attachment. However, the SOAP call did not define a short timeout, so an unavailable parser endpoint could still block the request until the PHP/network default timeout.

I will add a small defensive change in `ParseUtility::documentParse()` so the parser is skipped when SOAP is unavailable and the SOAP request uses a short connection/socket timeout. After that, parser unavailability should degrade to the existing non-parsed resume fallback instead of leaving the UI waiting on a long network timeout.

RussH:
Thanks, will wait for that update.

anonymoususer72041:
I’ve added the defensive parser fallback update now.

`ParseUtility::documentParse()` now skips parsing immediately if PHP SOAP support is unavailable. The SOAP client creation and the `DocumentParse()` call are both handled inside the same `try/catch` path and the parser request now uses a short explicit timeout via `connection_timeout` plus a temporary `default_socket_timeout` override. The previous socket timeout is restored afterwards.

With this in place, an unavailable external parser service should no longer leave the UI waiting on a long PHP/network default timeout. Parser unavailability now returns `false` and falls back to the existing non-parsed resume behavior rather than blocking candidate creation or resume handling.

RussH (approved):
Thanks — this looks good to me now. Good to merge.

PR #803, opened 2026-06-08
feat: detect and gate pending install migrations

anonymoususer72041:
## Summary

This PR adds automatic detection for pending install schema migrations and prevents OpenCATS from continuing normal operation while those migrations still need to be finalized.

It introduces a central, read-only migration status check for the install schema (`lib/SchemaMigrationStatus.php`). While install migrations are pending, it shows authenticated users a **Maintenance Required** notice instead of the normal UI, blocks application AJAX requests, and returns a maintenance message for the public Careers Portal.

Site administrators are pointed to the existing Installation Wizard (`installwizard.php`) to complete the upgrade, while recruiters and other non-admin users are shown a message telling them to contact their administrator.

The integration test setup finalizes the install schema version dynamically from `modules/install/Schema.php`, so CI test databases represent an already finalized installed snapshot without hard-coding the current install schema version in `db/cats_schema.sql`.

## Motivation

Previously, pending install schema migrations could be difficult to notice during normal application usage. This could allow users to continue using OpenCATS while the database schema was not fully aligned with the application code and the Careers Portal could remain publicly available during that state.

Detecting pending migrations centrally makes this state explicit and safer. Blocking the Careers Portal avoids exposing applicants to an application that still requires maintenance, while gating authenticated application requests prevents existing sessions or direct module URLs from bypassing the warning.

Linking administrators to the existing Installation Wizard keeps migration execution on the current, well-tested path and avoids introducing a second migration-execution path in this PR. Automating the migration itself is deliberately left to a separate follow-up PR so this change stays focused on detection and gating.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,8 @@ jobs:
           [ -n "$integrationtestdb_container" ] || { echo "integrationtestdb container ID not found"; exit 1; }
           timeout 180s bash -c "until [ \"\$(docker inspect --format '{{.State.Health.Status}}' \"${integrationtestdb_container}\" 2>/dev/null)\" = \"healthy\" ]; do sleep 3; done"
 
+          docker compose -f docker-compose-test.yml exec -T --workdir /var/www/public php php test/scripts/finalizeInstallSchemaVersion.php
+
           # 2. Pre-clean the integration database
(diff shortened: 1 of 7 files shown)


anonymoususer72041:
As this PR currently should not be merged, I will mark it as draft for now.

anonymoususer72041:
I've reduced this PR to detect and gate pending install migrations only. Authenticated users now see a maintenance notice, and administrators are linked to the existing Installation Wizard instead of a self-service maintenance action.
This PR no longer introduces a second migration-execution path.

The automatic migration itself will be implemented in a separate follow-up PR.

PR #804, opened 2026-06-09
fix: correct migration 377 regex escaping

anonymoususer72041:
This PR fixes the escaping of the regular expression used in migration 377.

The affected code is stored as a PHP string and evaluated later, so the regex needs to be escaped for both the surrounding migration string and the evaluated PHP code. The previous escaping caused the evaluated regex to differ from the intended pattern.

The updated version preserves the intended match behavior while making the evaluated regex valid and reliable. It also keeps the replacement string quoted correctly so `$2` is treated as the regex capture group replacement instead of a PHP variable.

### DIFF
--- a/modules/install/Schema.php
+++ b/modules/install/Schema.php
@@ -1476,8 +1476,8 @@ class CATSSchema
                         $notes = $row[\'notes\'];
 
                         $cleanedNotes = preg_replace(
-                            "/<span\\b(?=[^>]*\\bstyle\\s*=\\s*([\\\"\\\'])[^\\\"\\\']*\\bcolor\\s*:\\s*#ff6c00\\b[^\\\"\\\']*\\1)[^>]*>(.*?)<\\/span>/is",
-                            "$2",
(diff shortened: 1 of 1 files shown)


anonymoususer72041:
@lucquach I think this fix is the needed fix, isn't it?

anonymoususer72041:
@RussH please take a deeper look into this fix. I think this should solve a problem which was caused by this migration.

lucquach:
Hi @anonymoususer72041 

You are absolutely right! Thanks for catching this and putting up the PR.

When I originally provided the rewritten block for step 377 to fix the LIMIT 200 infinite loop/timeout issue, I messed up the regex escaping inside the eval() PHP string context. The [\\\"\\\'] sequence I wrote ended up terminating the string prematurely during eval(), causing a fatal syntax error.

Switching the regex to use single quotes and simplifying the escape sequence as you did is the perfect fix. Sorry for missing the notification earlier, and thank you again for cleaning up my mess!

RussH (approved):
Thank you — CI is passing, so I’m happy to approve.

The branch is out of date with master; I have updated it, lets see if it still passes.

PR #806, opened 2026-06-10
chore: add IANA timezone column to site schema

anonymoususer72041:
## Summary

This PR adds a new `time_zone_iana` column to the `site` table.

Fresh installations now create the column as part of the base schema and existing installations receive it through schema update. The column defaults to `UTC`.

This PR intentionally does not change the existing legacy `site.time_zone` column, installer behavior, runtime timezone handling, timestamp storage or timestamp display behavior.

## Motivation

OpenCATS currently relies on a legacy numeric timezone offset. This is not sufficient for future timezone-aware timestamp handling, especially if timestamp storage is later migrated toward UTC and display logic needs to use real regional timezone rules.

Before the installer or settings screens can start writing an IANA timezone identifier, the database schema must provide a place to store it. This is especially important for upgraded installations: the installer may run before later schema migrations or setup changes can depend on the new field. Without adding the column first, future installer or upgrade logic would not be able to reliably persist the selected IANA timezone for both fresh and existing instances.

This PR is therefore a small preparatory schema change. It makes `site.time_zone_iana` available while keeping all existing offset-based behavior unchanged. The actual transition to writing IANA timezone values, changing runtime timezone handling, migrating timestamps to UTC, ..., are intentionally left for follow-up PRs.

### DIFF
--- a/db/cats_schema.sql
+++ b/db/cats_schema.sql
@@ -992,6 +992,7 @@ CREATE TABLE `site` (
   `account_deleted` int(1) NOT NULL DEFAULT '0',
   `reason_disabled` text CHARACTER SET utf8,
   `time_zone` int(5) DEFAULT '0',
+  `time_zone_iana` varchar(64) COLLATE utf8_unicode_ci NOT NULL DEFAULT 'UTC',
   `time_format_24` int(1) DEFAULT '0',
   `date_format_ddmmyy` int(1) DEFAULT '0',
   `default_phone_country_code` varchar(8) COLLATE utf8_unicode_ci NOT NULL DEFAULT '+1',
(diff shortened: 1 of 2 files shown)


anonymoususer72041:
@RussH we want to merge this PR rather sooner than later as it needs to be included in a previous release regarding upcoming UTC work.

RussH (changes requested):
Can we change this to use plain `ALTER TABLE` rather than `ALTER IGNORE TABLE`?

`ALTER IGNORE TABLE` was deprecated/removed in MySQL 5.6/5.7, and we are only adding a new column, `IGNORE` is not needed.

Also, for consistency with `cats_schema.sql`, can we explicitly define the column collation in the migration as well, e.g. `COLLATE utf8_unicode_ci`?

anonymoususer72041:
Sorry, just saw your comment now.

Feel free to push into my PRs or suggest changes via GitHubs review function which I can directly approve via GitHub next time you find something.

> Also, for consistency with `cats_schema.sql`, can we explicitly define the column collation in the migration as well, e.g. `COLLATE utf8_unicode_ci`?

This should be superseeded by #805 anyways.

PR #807, opened 2026-06-10
fix: correct column position syntax in migration 380

anonymoususer72041:
This fixes the SQL syntax used in migration 380 when adding the `date_occurred` column to the `activity` table.

The migration previously used `BEFORE date_created`, which is not valid syntax for `ALTER TABLE ... ADD COLUMN` in MySQL/MariaDB. This could cause the migration to fail when upgrading an existing installation.

The column is now added `AFTER entered_by`, which is valid MySQL/MariaDB syntax and preserves the intended column order because `entered_by` is directly before `date_created` in the current schema.

This keeps the migration aligned with the schema used for fresh installations.

### DIFF
--- a/modules/install/Schema.php
+++ b/modules/install/Schema.php
@@ -1520,7 +1520,7 @@ class CATSSchema
             '380' => '
                 ALTER TABLE `activity`
                 ADD COLUMN `date_occurred` datetime NOT NULL DEFAULT \'1000-01-01 00:00:00\'
-                BEFORE `date_created`;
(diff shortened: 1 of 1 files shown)


anonymoususer72041:
@RussH this PR also needs to be merged before next release

RussH (changes requested):
Thanks again! Please can you update this branch by merging current `master` into it, then I’ll re-check CI and approve/merge.

RussH (approved):
have merged current master, all good now.

PR #809, opened 2026-06-14
fix: PHP 8.x compatibility - count() on null, mysqli_query() args, session_start() duplicate

ocjorge:
- AJAXInterface.php: Fix duplicate session_start() using session_status() check
- ZipLookup.php: Initialize variables before use, guard foreach against null, add isset() checks
- CareerPortal.php: Wrap PHPMailer send in try-catch to prevent fatal error when mail not configured
- candidates/Add.tpl: Add is_array() check before accessing parsingStatus array
- candidates/Search.tpl: Add isset() check for keySkills key
- candidates/Show.tpl: Cast extraFieldRS to array in count() calls
- candidates/Edit.tpl: Cast extraFieldRS to array in count() calls
- companies/Show.tpl: Add is_array() guards for contactsRS, departmentsRS; cast extraFieldRS
- companies/Edit.tpl: Cast extraFieldRS to array in count() calls
- companies/Add.tpl: Cast extraFieldRS to array in count() calls
- contacts/Show.tpl: Cast extraFieldRS to array in count() calls
- contacts/Edit.tpl: Cast extraFieldRS to array in count() calls
- contacts/Add.tpl: Cast extraFieldRS to array in count() calls
- joborders/Show.tpl: Cast extraFieldRS to array in count() calls
- joborders/Edit.tpl: Cast extraFieldRS to array in count() calls
- joborders/Add.tpl: Cast extraFieldRS to array in count() calls
- home/Home.tpl: Add is_array() guard for placedRS
- install/backupDB.php: Fix mysqli_query() argument order (PHP 8 changed parameter order)
- careers/CareersUI.php: Remove useCookie dependency for registration block display
- rss/index.php: Define LEGACY_ROOT constant if not already defined

Tested on PHP 8.4.21 + MariaDB 10.11 on Debian 12

### DIFF
--- a/lib/AJAXInterface.php
+++ b/lib/AJAXInterface.php
@@ -204,7 +204,7 @@ class SecureAJAXInterface extends AJAXInterface
         /* Give the session a unique name to avoid conflicts and start the
          * session. */
         @session_name(CATS_SESSION_NAME);
-        session_start();
+        if (session_status() === PHP_SESSION_NONE) { session_start(); }
 
         /* Validate the session. */
         if (!$this->isSessionLoggedIn())
--- a/lib/CareerPortal.php
+++ b/lib/CareerPortal.php
@@ -468,6 +468,8 @@ class CareerPortalSettings
             );
         } catch (Exception $e) {
             // Mail not configured or failed - log error for debugging
+            $mailerStatus = false;
+
             error_log('OpenCATS Mailer error (site=' . $this->_siteID . '): ' . $e->getMessage());
         }
     }
--- a/lib/CareerPortal.php
+++ b/lib/CareerPortal.php
@@ -458,13 +458,17 @@ class CareerPortalSettings
 
         /* Send e-mail notification. */
         //FIXME: Make subject configurable.
-        $mailer = new Mailer($this->_siteID, $userID);
-        $mailerStatus = $mailer->sendToOne(
-            array($destination, ''),
-            $subject,
-            $body,
-            true
-        );
+        try {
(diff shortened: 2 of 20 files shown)


anonymoususer72041:
Thank you for your work, @ocjorge.

Have you seen #792? You might want to join this PR and comment outstanding work there.

ocjorge:
Thank you for the feedback and for pointing me to #792!

I reviewed #792 and I can see it already addresses PHP 8.5 compatibility 
comprehensively. My PR #809 was developed and tested on PHP 8.4.21 on 
Debian 12 (Bookworm) in a production-like environment as part of an 
academic server administration project.

Some fixes in #809 that may complement #792 and are not immediately 
obvious from the title:

- **CareerPortal.php**: Wraps PHPMailer `send()` in try/catch with 
  `error_log()` to prevent a fatal 500 error when no mail server is 
  configured — this affects the Career Portal candidate application flow.
- **AJAXInterface.php**: Guards `session_start()` with `session_status() 
  === PHP_SESSION_NONE` to prevent duplicate session warnings in AJAX calls.
- **ZipLookup.php**: Defensive guards for null/false return from 
  `simplexml_load_file()`, initialization of `$loc_level_*` variables 
  before use, and `rawurlencode()` on the zip parameter.
- **backupDB.php**: Fixes `mysqli_query()` argument order which changed 
  in PHP 8.

I am happy to:
1. Close #809 if these fixes are already covered in #792
2. Cherry-pick specific commits into #792 if any are useful
3. Continue iterating on #809 independently

Whatever works best for the project. Thank you for maintaining OpenCATS!

PR #811, opened 2026-06-14
refactor: add Laravel wrapper for legacy application

anonymoususer72041:
This PR introduces the first step towards gradually moving OpenCATS to Laravel. Instead of changing the existing application behavior directly, the current codebase is placed behind a new Laravel application shell so future migration work can happen incrementally.

The legacy application remains responsible for the existing functionality, while Laravel becomes the new outer application structure and public entry point. This creates a foundation for moving individual parts of OpenCATS into Laravel over time without requiring a disruptive full rewrite.

The existing legacy code is moved into a dedicated `legacy/` directory and the new wrapper keeps the current entry points available through Laravel. Supporting project structure, configuration, development tooling and runtime defaults are added so the repository can start evolving as a Laravel-based application.

This change is intended as an architectural foundation only. It does not aim to replace legacy functionality yet, but prepares the codebase for smaller, safer follow-up PRs that can migrate features step by step.


anonymoususer72041:
#792 needs to be merged first to continue work on this PR.

PR #812, opened 2026-06-16
feat: add configurable 12-hour and 24-hour time formats

anonymoususer72041:
## Summary

This PR adds a configurable site-wide time format setting that allows OpenCATS installations to use either 12-hour or 24-hour time display.

This wires the existing `site.time_format_24` setting into site localization, session handling, localization settings, the first-run wizard and the installer. It also applies the configured time format across activity display and editing, calendar forms and settings, email template date/time replacement and other user-facing date/time output where appropriate.

The default behavior remains unchanged: existing installations continue to use the 12-hour format unless the 24-hour format is explicitly selected.

## Motivation

This improves OpenCATS internationalization by allowing installations to use the time format that matches their locale and users' expectations.

A large share of countries commonly use the 24-hour time format. Supporting both 12-hour and 24-hour formats makes OpenCATS more suitable for international deployments while preserving the existing default behavior for users who prefer the 12-hour format.

### DIFF
--- a/ajax/editActivity.php
+++ b/ajax/editActivity.php
@@ -80,7 +80,6 @@ $activityNote = trim(urldecode($_POST['notes']));
 $activityDate = trim(urldecode($_POST['date']));
 $activityHour = trim(urldecode($_POST['hour']));
 $activityMinute = trim(urldecode($_POST['minute']));
-$activityAMPM = trim(urldecode($_POST['ampm']));
 
 $dateFormatFlag = $_SESSION['CATS']->isDateDMY()
     ? DATE_FORMAT_DDMMYY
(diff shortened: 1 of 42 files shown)


RussH (changes requested):
PR #812 — Configurable 12-hour / 24-hour time formats


    In activity.js, can you re-check the append logic around the date/hour/minutte/eetc fields? It looked like some of the old append code and new conditional append code may both still be present?

anonymoususer72041:
I re-checked the `activity.js` append logic around the date/time controls.

In the current branch state, I only see the date, hour and minute controls being appended once. The AM/PM select is created and appended only conditionally for 12-hour mode and the submit path also avoids reading the AM/PM field when it does not exist in 24-hour mode.

I do not see a remaining duplicate append path in the current diff. Could you point me to the specific lines you had in mind if you are still seeing one?

PR #845, opened 2026-08-29
fix: resolve relative SystemUtility includes in installer AJAX endpoint

ocjorge:
## Problem

The `resumeParsing` case in `modules/install/ajax/ui.php` contained 
four repeated relative `include_once('lib/SystemUtility.php')` calls 
inside individual `else` blocks. Relative includes can fail when the 
endpoint is accessed directly from a different working directory.

## Solution

Replace the four repeated relative includes with a single 
`include_once(LEGACY_ROOT . '/lib/SystemUtility.php')` at the top of 
the `resumeParsing` case, before the conditional blocks.

This also removes the redundancy — `include_once` only loads the file 
once anyway, so four calls in separate branches were unnecessary.

## Context

This was noted as a pre-existing issue in PR #797:
> "modules/install/ajax/ui.php still contains some pre-existing 
> relative includes such as include_once('lib/SystemUtility.php') 
> in the resume parsing path. Those lines were already present before 
> this branch and were not introduced by this commit. Since they are 
> a separate legacy relative-path issue rather than part of the 
> bootstrap-order change, I would prefer to handle them in a follow-up PR."

### DIFF
--- a/modules/install/ajax/ui.php
+++ b/modules/install/ajax/ui.php
@@ -293,6 +293,8 @@ switch ($action)
     case 'resumeParsing':
         echo '<script type="text/javascript">setActiveStep(4);</script>';
 
+        include_once(LEGACY_ROOT . '/lib/SystemUtility.php');
+
         if (ANTIWORD_PATH == '')
         {
             echo '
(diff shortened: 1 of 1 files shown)


RussH:
@ocjorge Excellent PR — very clean, focused change and exactly the sort of tidy-up we need. Everything looks good from my side. I’ll just update it against current `master`, let CI run again, and then merge.

Thank you for this — much appreciated. I look forward to the next one!

RussH:
@ocjorge  thanks again, I'm just pulling together an OpenCATS roadmap - will add it as a project, please  review it, see if anything there looks good to work on next?

PR #863, opened 2026-09-02
fix: report the no-log success status from the queue CLI

darkdi:
The status switch at the end of the script lists TASKRET_SUCCESS twice, and the second branch prints "SUCCESS (NO LOG)", which is the wording for TASKRET_SUCCESS_NOLOG. QueueProcessor::startTask returns whatever the task returned, so a task finishing without a log currently leaves the status line blank. I read this off constants.php and QueueProcessor, I have not run the queue processor itself.

### DIFF
--- a/QueueCLI.php
+++ b/QueueCLI.php
@@ -101,7 +101,7 @@ switch($retVal)
     case TASKRET_SUCCESS:
         echo "SUCCESS";
         break;
-    case TASKRET_SUCCESS:
+    case TASKRET_SUCCESS_NOLOG:
         echo "SUCCESS (NO LOG)";
         break;
 }


RussH:
@darkdi Excellent catch, thank you.

A very  small and focussed fix - exactly the sort of small issue that is easy to overlook in old code. 

If you fancy picking up another one, please have a look at the **Projects → Modernisation roadmap**. There are quite a few small, well-contained fixes and cleanup items there, and contributions in this style are exactly what we’re looking for. It would be great to see another PR from you!

darkdi:
Will do, thanks. I had a second one from the same pass sitting unsent: in lib/Candidates.php the EEO defaults array lists veteranTracking twice in a row, both times '0'. Nothing behaves differently, the line is just dead, so it can wait for a batch if you would rather not spend a review on it.

Issue #52, opened 2016-02-10
Mismatch bei Date format.


prolyfix:
Hi,
(my first comment on GitHub). I Installed OpenCATS yestedays using the date format (DD-MM-YY)
when I create activity and want later to modify it, I have a date mismatch:
Instead of giving me the 10. February .2016 it gives me as Oktober, 2 2016

RussH:
Hi, I believe some recent PHP changes mean that if your server timezone isn't the same as your target timezone, you'll have issues. I need to check through the code to see why this is happening - however the quick fix it to have your server time set to the same as your required timezone. Clearly if you're installing on shared hosting this will be a problem until resolved.

filippo-basso:
Hi,
seems I've the same problem, when editing an activity that is dated 10-4-2017 (date format DD-MM-AAAA) it is presented below as 4 October 2017, and then if not changed manually, after saving the change it will present 4-10-2017 as activity date.

**Temporary workaround:** use the MM-DD-AAAA date format (but it's not too intuitive for several countries...)

Technically, seems not related with the server timezone, seems:
```
# cat /etc/timezone
Etc/UTC
```

RussH:
Closing - duplicate of #344 (also a fix is included in that issue)

PR #876, opened 2026-09-17
feat: add YYYY-MM-DD date format option

rapsealk:
## Summary

This PR adds a third site-wide date format, **YYYY-MM-DD (ISO 8601)**, alongside the existing MM-DD-YYYY (US) and DD-MM-YYYY (UK) options.

The option is available in Settings → Administration → Localization, the installer and the new-install wizard. When selected, it applies to displayed dates (SQL `DATE_FORMAT` output), date pickers, form validation and parsing, date extra fields (display, edit, data grid and export), calendar add/edit, email template date variables and the job order report period line.

The default behavior is unchanged: existing installations keep their current format.

## Motivation

YYYY-MM-DD is the standard or common date format in many locales (for example Korea, Japan, China, Sweden, Lithuania and Canada), and it is unambiguous for international teams. The current D-M-Y/M-D-Y boolean doesn't allow it.

## Implementation

- **No schema migration.** `site.date_format_ddmmyy` (`INT(1)`) now stores `0` (M-D-Y), `1` (D-M-Y) or `2` (Y-M-D). Existing `0`/`1` values keep their meaning.
- `CATSSession::getDateFormat()` returns the existing `DATE_FORMAT_MMDDYY` / `DATE_FORMAT_DDMMYY` / `DATE_FORMAT_YYYYMMDD` flag. `isDateDMY()` is kept for existing callers.
- `Site::setLocalization()` and `CATSSession::setTimeDateLocalization()` accept either a date format flag or the previous boolean D-M-Y value (`DateUtility::normalizeDateFormat()`).
- Sessions created before the upgrade keep their date format: the old `_dateDMY` property is still declared and is migrated to `_dateFormat` in `CATSSession::__wakeup()`.
- Dashboard weekly graph labels use `d/m` (D-M-Y), `m-d` (Y-M-D) or `m/d` (M-D-Y).
- New `DateUtility` helpers replace the repeated `isDateDMY() ? … : …` ternaries:
  - `getDateFormatFromSiteValue()`, `getSiteValueFromDateFormat()`, `getDateFormatFromFormValue()`
  - `getJsDateFormat()` / `getPhpDateFormat()` (default to the session format, like `getTimeFormat()`)
  - `formatStoredDate()` converts extra field dates, which are still stored as `MM-DD-YY`, for display.
- `DatabaseConnection` rewrites `%m-%d-%y` / `%m/%d/%Y` etc. to `%Y-%m-%d` / `%Y/%m/%d` in Y-M-D mode, just as it already does for D-M-Y.
- Y-M-D uses a 4-digit year, which `DateUtility::validate()`/`convert()` (`DATE_FORMAT_YYYYMMDD`) and `calendarDateInput.js` (`YYYY-MM-DD`) already support.

## Testing

- Unit tests: added `testDateFormatMapping` and `testFormatStoredDate` to `DateUtilityTest`, and `SessionDateFormatTest` (boolean setter values, migration of pre-upgrade sessions). `UnitTests` pass (124 tests) on PHP 8.5.
- Integration tests: `IntegrationTests` pass (20 tests) against `docker-compose-test.yml`.
- Manual checks with curl against the `docker-compose-test.yml` stack, logged in as `testerRoot`:
  - Localization settings: D-M-Y was preselected. Saving `ymd` stored `date_format_ddmmyy = 2`, and after logging back in `YYYY-MM-DD (ISO)` was preselected.
  - Candidate add page emitted `CATSUserDateFormat = 'YYYY-MM-DD'`. Submitting `dateAvailable=2026-09-30` stored `2026-09-30`. Submitting `09-30-26` was rejected with "Invalid availability date".
  - Candidate show/edit pages and the calendar date input render `YYYY-MM-DD`.
  - Date extra field: `2026-10-01` was stored as `10-01-26` (unchanged storage format), shown as `2026-10-01` on show/edit, and the data grid `pagerRender`/`exportRender` output was `2026-10-01`.
- Browser tests (Playwright, Chromium) in each of M-D-Y, D-M-Y and Y-M-D. All dates use days above 12, so a day/month swap would show:
  - Calendar add: clicking Oct 23 prefilled `10-23-26` / `23-10-26` / `2026-10-23`. DB stored `2026-10-23`, and the event panel showed the same format.
  - Calendar edit: the form prefilled the event date. Moving it to the 27th stored `2026-10-27`.
  - Activity inline edit (`js/activity.js` → `ajax/editActivity.php`, the #344/#52 path): Oct 14 → Oct 27 was displayed correctly before and after a reload, and DB stored `2026-10-27`.
  - Dashboard weekly labels: `09/14 - 09/20` / `14/09 - 20/09` / `09-14 - 09-20`.
  - Session upgrade: with the site in D-M-Y, a session file rewritten to the old shape (`_dateDMY` = true) still rendered `DD-MM-YY`, was re-saved with `_dateFormat` = D-M-Y and `_dateDMY` = null, and logged no deprecation.

## Notes

- The dashboard weekly graph image labels its x-axis `0`…`14` (artichow), so the week labels are not visible in the rendered image. This is not changed by this PR, and I have not checked whether `master` does the same.
- Unrelated issue found while testing (present on `master` too): date-only values such as `date_available` are shifted by the time zone offset in `DATE_FORMAT` output (e.g. `2026-09-30` shown as `09-29-26` with `OFFSET_GMT=2` and site time zone 0).

## Related

- #344 and #52 (closed): day and month were swapped when editing activities in D-M-Y mode. Y-M-D goes through the same activity date parse/edit path (`js/activity.js` → `DateInputForDOM`, `ajax/editActivity.php`), so that flow is worth checking in review.
- No existing issue requests a Y-M-D / ISO 8601 date format.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RussH (changes requested):
Thanks for contributing this — YYYY-MM-DD support is a useful addition to OpenCATS, and I really appreciate you taking the time to work through the existing date-handling paths rather than bolting on a separate implementation. I’d definitely like to see more contributions from you.

I am being a little pickier than usual at the moment, mainly because I’ve had to do some cleanup after a few recent merges, so I’d appreciate it if you could accommodate a few changes/enhancements before I approve this one:

Session compatibility: replacing the declared _dateDMY property means existing serialized sessions can produce a PHP dynamic-property deprecation and temporarily fall back to M-D-Y. Please preserve compatibility for users who are already logged in rather than relying on a logout/login after upgrade.
Dashboard: Dashboard.php still treats the format as DMY vs everything else, so YYYY-MM-DD currently falls back to M/D graph labels. Please update this to handle the new format properly.
Setter compatibility: Site::setLocalization() and CATSSession::setTimeDateLocalization() previously accepted a boolean DMY value. Please retain backward compatibility for existing callers while also supporting the new date-format constants.
Browser coverage: please give the actual browser paths a quick test, particularly Calendar add/edit and Activity inline edit. These are the areas I’m most cautious about because date parsing there has caused issues historically.

One integration point to be aware of: there is quite a lot of overlap with the Bootstrap 5.3 conversion currently underway in feature/bootstrap-5.3, including Calendar, Candidates, Contacts, Job Orders and the shared Activity modal. That’s not a blocker for this PR — just please keep the changes focused on the date-format functionality and avoid unrelated markup/template cleanup. I can then carry the date-format changes into the Bootstrap-converted versions when integrating it.

I haven’t found any obvious new security concerns, and overall I think the direction of the implementation is good. If you can address the points above, I should be much happier approving it.

Thanks again for taking the time to contribute this.

rapsealk:
@RussH Thanks for the detailed comments! I’ll work through the points you mentioned and polish the PR accordingly. No rush on the review, so please take your time!

rapsealk:
@RussH Thanks again for the review. I've pushed one commit per point:

- **Session compatibility** (92c45dd): `_dateDMY` is declared again, so there's no dynamic-property deprecation, and `CATSSession::__wakeup()` migrates it to `_dateFormat`. Users who are logged in during the upgrade keep D-M-Y without logging in again. I tested this with a real session file rewritten to the old shape, and there's a unit test for it too.
- **Dashboard** (de0b765): weekly labels now use `d/m` for D-M-Y, `m-d` for Y-M-D and `m/d` for M-D-Y. Y-M-D renders as `09-14 - 09-20`. If `m/d` or `Y-m-d` reads better to you, I can switch it.
- **Setter compatibility** (dfc3c2e): `Site::setLocalization()` and `CATSSession::setTimeDateLocalization()` accept the old boolean again, as well as the date format constants. This also fixes a bug in the previous version: `true` was silently stored as M-D-Y, because the constants are 1/2/4 and `match` compares strictly.
- **Browser coverage**: I tested Calendar add, Calendar edit and Activity inline edit in Chromium with Playwright, in all three formats, using days above 12 so a day/month swap would show. Every check passed, and the details are in the PR description.

The template changes are still limited to the date-format lines, so they should carry over to the Bootstrap 5.3 versions easily.

One thing I noticed that isn't changed here: the weekly dashboard graph image labels its x-axis `0`…`14`, so the week labels don't show in the rendered image. I haven't checked whether `master` does the same.

